PR #15, opened 2023-06-23
add send message and display features

Djokovic0311:


components/chat/Chat.tsx
```diff
@@ -1,10 +1,37 @@
...
+    setMessages((messages) => [
+      ...messages,
+      { user: currentUser, message: messageInput, timestamp: new Date() },
+    ]);
+    // clear the input field
+    setMessageInput('');
+  };
+  const handleUserChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setCurrentUser(event.target.value);
+  };
```
Raduc4, line 23: good we will need to refactor this to use React context in the future

PR #47, opened 2024-08-03
[KAN-2] UI Redesign

kyle-tennison:
To help with the redesign, I moved @Raduc4 's code to `.archive` and slowly pulled features back in. At this point, just about everything that was implemented is reimplemented, with exception to notification handling (which won't be  hard to pull back in).


src-tauri/src/commands/mod.rs
```diff
@@ -1,36 +1,96 @@
...
+    let mut guard = state.listeners.lock().await;
+    guard.push(packet_handle);
+    drop(guard);
+
+    // Wait for the background TCP listener (main.rs) to dispatch the message
+    let incoming = match rx.recv().await {
+        Some(v) => v,
+        None => panic!("Channel unexpectedly closed before response."),
+    };
+
```
tbraun96: Store the packet handle *first*. Otherwise, you allow the possibility for this racey scenario:

1. You send a packet
2. The IS responds almost instantly
3. the inbound packet gets routed, but since we haven't stored the transmitter/packet-handle yet, the packet has nowhere to go
4. Then, you register the packet_handle

It's not clear how we're handling packets that passively arrive. If this code works for now, great, however, this will all change once the PR is merged: https://github.com/Avarok-Cybersecurity/citadel-internal-service/pull/40

Your logic will simplify greatly, as you will just use the new Messenger layer.

src-tauri/src/structs.rs
```diff
@@ -1,11 +1,23 @@
...
 use citadel_internal_service_types::InternalServiceResponse;
 use serde::{Deserialize, Serialize};
+use tokio::sync::mpsc::Sender;
 use tokio::sync::Mutex;
+use uuid::Uuid;
+
+pub struct PacketHandle {
+    pub request_id: Uuid, // The ID to listen for in the response stream
+    pub channel: Sender<InternalServiceResponse>, // The channel to stream the response to
+}
```
tbraun96, line 15: Your intuition for handling sending and receiving is generally correct. I basically do the same thing in this PR: https://github.com/Avarok-Cybersecurity/citadel-internal-service/pull/40

tbraun96 (approved):

PR #59, opened 2026-07-12
fix(server-kernel): seed workspace structure only into a fresh workspace

Tjemmmic:
## Overview

`initialize_workspace_structure` is called unconditionally on every `on_start`, but it mints fresh `Uuid::new_v4()` office/room IDs and *appends* them to the existing node set. It has no already-seeded guard — unlike `inject_admin_user` directly above it, which correctly guards on `workspace_exists`.

Against a persistent (filesystem) backend, this means **every server restart deposits another complete copy of the entire office/room tree**. Since `deploy.sh` restarts the server on every deploy, a live workspace corrupts itself a little more with each update, and the damage is written to disk with new UUIDs — it does not self-heal.

Reproduced on real hardware with `WORKSPACE_BACKEND=filesystem`:

- Boot 1 (fresh volume): 3 offices, 8 rooms created (the baked-in `General` / `Tutorials` / `Welcome`)
- Boot 2 (restart, same volume): **6 offices, 16 rooms**

## Why this was never caught

Dev and CI run the server on the **in-memory** backend, where every boot legitimately starts from an empty store, so seeding-on-every-boot is indistinguishable from correct behavior. Only `docker-compose.production.yml` sets `WORKSPACE_BACKEND=filesystem`. The `production-build` CI job green-lights an image whose persistence path is never booted a second time, so this class of bug can only surface in production.

## Fix

Return early from `initialize_workspace_structure` when any node is already parented to the workspace root.

The detector is not new: `parent_id.as_deref() == Some(WORKSPACE_ROOT_ID)` is verbatim the predicate the codebase already uses to enumerate top-level offices in `async_node_ops.rs:216`. Runtime-created offices pass `WORKSPACE_ROOT_ID` as their parent (`async_node_ops.rs:86`), and the root node itself carries `parent_id: None` (`async_node_ops.rs:222`) — so the guard correctly detects user-created trees and correctly excludes root.

## Design notes

**Skip rather than upsert.** The office tree is user-mutable after the first boot: offices are created and renamed, MDX content is edited live. The baked-in `content_base_dir` describes the *initial* state of a new workspace, not a desired state to reconcile toward. Re-applying config on every boot would clobber user edits even if it didn't duplicate.

**Behavior change worth calling out.** Editing `documents/defaults/root/**/CONTENT.md` and redeploying no longer alters an existing workspace. It never usefully did — it produced duplicate offices — but the config is now explicitly initial-state-only.

**Accepted edge case.** If an admin deletes *every* office, a restart re-seeds the defaults. A persistent sentinel (like the existing `KEY_MIGRATION_DONE`) would prevent that, but the children-check self-heals if a seed ever fails midway and adds no persistent state. This felt like the better trade.

## Tests

Two regression tests, both proven non-vacuous (they fail when the guard is removed, reproducing `2 nodes -> 4 nodes`):

- `second_seed_does_not_duplicate_the_tree` — seeds twice against one store (the unit-test equivalent of a restart) and asserts the node set is unchanged. It pins **node identity**, not just cardinality: a re-seed that wiped and recreated the tree would keep the count equal while silently rotating every UUID and dangling every stored reference.
- `existing_user_created_office_suppresses_seeding` — a workspace whose defaults were renamed or replaced must never have them re-injected.

## Verification

- `cargo test -p citadel-workspace-server-kernel`: 145 passed, 0 failed
- `cargo clippy -p citadel-workspace-server-kernel -- -D warnings`: clean
- `cargo fmt -p citadel-workspace-server-kernel --check`: clean
- End-to-end on the deploy box with the filesystem backend: rebuilt the server image with this fix and ran **3 consecutive restarts** against one volume — offices stayed at 3, rooms at 8, and the skip guard fired on each restart.

## No regression to dev/CI

The dev compose leaves `WORKSPACE_BACKEND` unset, so the server runs in-memory and every boot starts with an empty store. `get_all_nodes()` returns empty, the guard does not fire, and seeding happens exactly as before. Integration tests that restart backend services are unaffected.

## Follow-up (not in this PR)

CI has no job that boots the server twice against a filesystem backend. That gap is precisely why this shipped, and closing it would catch this entire class of bug in minutes.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -499,6 +499,51 @@ jobs:
           fi
           echo "UI is serving the SPA shell"
           echo "Smoke test passed: server, internal-service, ui all up"
+      - name: Persistence - restarting the server must not duplicate the workspace tree
+        # THE gap that let the structure-duplication bug ship. Every other test runs the server
+        # on the IN-MEMORY backend, where each boot legitimately starts from an empty store - so
+        # seeding-on-every-boot is indistinguishable from correct behaviour and no unit test can
+        # see the bug. Only a real restart against a real filesystem backend exposes it, and
+        # `docker-compose.production.yml` is the only place the server runs that way.
+        #
+        # The server container from the smoke test above is already running with
+        # WORKSPACE_BACKEND=filesystem on the `server_data` volume. Restart it against that SAME
+        # volume and assert it did not seed a second copy of the office/room tree.
+        env:
+          WORKSPACE_MASTER_PASSWORD: 'ci-smoke-test'
+          INTERNAL_SERVICE_PORT: '12345'
+        run: |
+          set -euo pipefail
+          compose="docker compose -f docker-compose.production.yml"
+
+          offices_before=$($compose logs server 2>&1 | grep -c "Creating office node" || true)
+          rooms_before=$($compose logs server 2>&1 | grep -c "Creating room node" || true)
+          echo "Before restart: ${offices_before} office / ${rooms_before} room creations."
+
+          if [ "$offices_before" -eq 0 ]; then
+            echo "::error::The server seeded no offices at all, so this check would be vacuous."
+            $compose logs server 2>&1 | tail -40
+            exit 1
+          fi
+
+          echo "Restarting the server against the same data volume..."
+          $compose restart server
+          timeout 180 bash -c 'until docker compose -f docker-compose.production.yml ps server --format json | grep -qw healthy; do sleep 5; done' || {
+            echo "::error::Server did not come back healthy after restart"
+            $compose logs server 2>&1 | tail -40
+            exit 1
+          }
+
+          offices_after=$($compose logs server 2>&1 | grep -c "Creating office node" || true)
+          rooms_after=$($compose logs server 2>&1 | grep -c "Creating room node" || true)
+          echo "After restart:  ${offices_after} office / ${rooms_after} room creations."
+
+          if [ "$offices_after" -ne "$offices_before" ] || [ "$rooms_after" -ne "$rooms_before" ]; then
+            echo "::error::Restart re-seeded the workspace: offices ${offices_before} -> ${offices_after}, rooms ${rooms_before} -> ${rooms_after}. On a persistent backend this grows an extra copy of the whole tree on EVERY deploy."
+            $compose logs server 2>&1 | grep -E "Creating (office|room) node" | tail -20
+            exit 1
+          fi
+          echo "PASS: restart added no duplicate offices or rooms."
       - name: Dump Logs on Failure
         if: failure()
         run: |
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -516,17 +516,23 @@ jobs:
           set -euo pipefail
           compose="docker compose -f docker-compose.production.yml"
 
-          offices_before=$($compose logs server 2>&1 | grep -c "Creating office node" || true)
-          rooms_before=$($compose logs server 2>&1 | grep -c "Creating room node" || true)
-          echo "Before restart: ${offices_before} office / ${rooms_before} room creations."
-
-          if [ "$offices_before" -eq 0 ]; then
+          # Guard against a vacuous pass: if the FIRST boot seeded nothing, "the restart created
+          # nothing" would be trivially true and would prove nothing.
+          seeded_on_first_boot=$($compose logs server 2>&1 | grep -c "Creating office node" || true)
+          echo "First boot seeded ${seeded_on_first_boot} offices."
+          if [ "$seeded_on_first_boot" -eq 0 ]; then
             echo "::error::The server seeded no offices at all, so this check would be vacuous."
             $compose logs server 2>&1 | tail -40
             exit 1
           fi
 
-          echo "Restarting the server against the same data volume..."
+          # Only inspect logs emitted AFTER the restart. Comparing cumulative counts before/after
+          # would couple the result to docker's log buffering and rotation: if the daemon dropped
+          # early lines, the "after" count could come out LOWER and the check would misfire. A
+          # `--since` window asks the precise question instead - "did this boot create any office
+          # or room nodes?" - and the answer must be no.
+          restart_at=$(date -u +%Y-%m-%dT%H:%M:%SZ)
+          echo "Restarting the server against the same data volume (window starts ${restart_at})..."
           $compose restart server
           timeout 180 bash -c 'until docker compose -f docker-compose.production.yml ps server --format json | grep -qw healthy; do sleep 5; done' || {
             echo "::error::Server did not come back healthy after restart"
@@ -534,16 +540,24 @@ jobs:
             exit 1
           }
 
-          offices_after=$($compose logs server 2>&1 | grep -c "Creating office node" || true)
-          rooms_after=$($compose logs server 2>&1 | grep -c "Creating room node" || true)
-          echo "After restart:  ${offices_after} office / ${rooms_after} room creations."
+          created_after_restart=$($compose logs server --since "$restart_at" 2>&1 | grep -cE "Creating (office|room) node" || true)
+          echo "Nodes created by the restarted server: ${created_after_restart}"
 
-          if [ "$offices_after" -ne "$offices_before" ] || [ "$rooms_after" -ne "$rooms_before" ]; then
-            echo "::error::Restart re-seeded the workspace: offices ${offices_before} -> ${offices_after}, rooms ${rooms_before} -> ${rooms_after}. On a persistent backend this grows an extra copy of the whole tree on EVERY deploy."
-            $compose logs server 2>&1 | grep -E "Creating (office|room) node" | tail -20
+          if [ "$created_after_restart" -ne 0 ]; then
+            echo "::error::The restart re-seeded the workspace: it created ${created_after_restart} office/room nodes on top of the existing tree. On a persistent backend this grows an extra copy of the whole tree on EVERY deploy."
+            $compose logs server --since "$restart_at" 2>&1 | grep -E "Creating (office|room) node" | head -20
             exit 1
           fi
-          echo "PASS: restart added no duplicate offices or rooms."
+
+          # Positive confirmation, not just an absence: the guard must actually have run and said
+          # so. Without this, a server that failed to reach the seeding code at all would also
+          # produce zero creations and pass.
+          if ! $compose logs server --since "$restart_at" 2>&1 | grep -q "already seeded; skipping"; then
+            echo "::error::The restarted server never reported skipping the seed. It may not have reached the guard at all - this check cannot confirm the behaviour."
+            $compose logs server --since "$restart_at" 2>&1 | tail -30
+            exit 1
+          fi
+          echo "PASS: the restart created no offices or rooms, and the seed-once guard fired."
       - name: Dump Logs on Failure
         if: failure()
         run: |
--- a/citadel-workspace-server-kernel/src/kernel/async_kernel.rs
+++ b/citadel-workspace-server-kernel/src/kernel/async_kernel.rs
@@ -312,10 +312,20 @@ impl<R: Ratchet + Send + Sync + 'static> AsyncWorkspaceServerKernel<R> {
             //
             // So the marker goes first. A pending marker with no workspace is a harmless, self-
             // healing state; a workspace with no marker is a permanent one.
-            self.domain_operations
-                .backend_tx_manager
-                .mark_structure_seed_pending()
-                .await?;
+            //
+            // Armed ONLY when there is actually a structure to seed. `on_start` calls
+            // `initialize_workspace_structure` only when a structure is configured, so arming it
+            // unconditionally would leave a permanently-pending marker on a server booted without
+            // one. A later deploy that ADDED a structure config would then find `pending == true`
+            // against a workspace that has been live for weeks, and inject the baked-in defaults
+            // into it - an established workspace, which is exactly what this guard exists to
+            // prevent. The obligation only exists if something is going to discharge it.
+            if self.workspace_structure.is_some() {
+                self.domain_operations
+                    .backend_tx_manager
+                    .mark_structure_seed_pending()
+                    .await?;
+            }
 
             // Debug: Check current storage mode
             if self.domain_operations.backend_tx_manager.is_test_mode() {
@@ -630,13 +640,21 @@ impl<R: Ratchet + Send + Sync + 'static> AsyncWorkspaceServerKernel<R> {
             .values()
             .any(|node| node.parent_id.as_deref() == Some(crate::WORKSPACE_ROOT_ID))
         {
+            // A tree exists while a seed is still owed: the previous boot wrote the nodes but died
+            // before recording that it had. Finish that transition rather than half-completing it
+            // - mark it seeded AND discharge the obligation, so no dangling `pending` is left
+            // behind for a future reader to puzzle over.
             warn!(
                 target: "citadel",
-                "Workspace reported as new but already has a tree; refusing to seed over it"
+                "A seed is owed but the workspace already has a tree; recording it as seeded rather than seeding over it"
             );
             self.domain_operations
                 .backend_tx_manager
                 .mark_structure_seeded()
+                .await?;
+            self.domain_operations
+                .backend_tx_manager
+                .clear_structure_seed_pending()
                 .await?;
             return Ok(());
         }
@@ -837,12 +855,24 @@ impl<R: Ratchet + Send + Sync + 'static> AsyncWorkspaceServerKernel<R> {
             .mark_structure_seeded()
             .await?;
         // The debt is paid. `seeded` already takes precedence over `pending` on every later boot,
-        // so this is hygiene rather than correctness - it keeps the store from carrying a
-        // permanently-true "pending" flag that a future reader would have to reason about.
-        self.domain_operations
+        // so clearing it is hygiene, not correctness - it keeps the store from carrying a
+        // permanently-true flag that a future reader would have to reason about.
+        //
+        // Deliberately BEST-EFFORT. The tree is durably written and the workspace is durably
+        // marked seeded; failing the boot now over a cosmetic write would turn a completed seed
+        // into a crash loop, and the next boot would short-circuit on `seeded` anyway.
+        if let Err(e) = self
+            .domain_operations
             .backend_tx_manager
             .clear_structure_seed_pending()
-            .await?;
+            .await
+        {
+            warn!(
+                target: "citadel",
+                "Workspace seeded successfully, but could not clear the seed-pending marker: {e}. \
+                 Harmless - the seeded marker takes precedence on every later boot."
+            );
+        }
 
         info!(target: "citadel", "Workspace structure initialization complete");
         Ok(())
@@ -1495,7 +1525,19 @@ mod structure_seed_idempotency_tests {
 
     type Kernel = AsyncWorkspaceServerKernel<StackedRatchet>;
 
+    /// A kernel wired the way a real server is: it HAS a structure configured, so
+    /// `inject_admin_user` takes on a seed obligation when it creates a fresh workspace. Setting
+    /// the field directly (rather than going through the async constructor) keeps the tests off
+    /// the filesystem while still exercising the production wiring.
     fn kernel() -> Kernel {
+        let mut k = AsyncWorkspaceServerKernel::<StackedRatchet>::new(None);
+        k.workspace_structure = Some((one_office_one_room(), None));
+        k
+    }
+
+    /// A kernel with NO structure configured, as a server whose kernel.toml sets no content
+    /// directory would be.
+    fn kernel_without_structure() -> Kernel {
         AsyncWorkspaceServerKernel::<StackedRatchet>::new(None)
     }
 
@@ -1825,6 +1867,37 @@ mod structure_seed_idempotency_tests {
         assert!(
             backend.is_structure_seeded().await.unwrap(),
             "the workspace should now be marked seeded"
+        );
+    }
+
+    /// A server booted with NO structure configured must not arm a seed obligation.
+    ///
+    /// `on_start` only calls `initialize_workspace_structure` when a structure is configured, so
+    /// an obligation armed on such a boot would never be discharged - it would sit there while the
+    /// workspace ran and accumulated real content, and the first deploy that ADDED a structure
+    /// config would find `pending == true` and inject the baked-in defaults into an established
+    /// workspace. The obligation may only be taken on if something is going to discharge it.
+    #[tokio::test]
+    async fn no_structure_config_means_no_seed_obligation() {
+        let k = kernel_without_structure();
+        assert!(
+            k.workspace_structure.is_none(),
+            "precondition: this kernel has no structure configured"
+        );
+
+        k.inject_admin_user(MASTER_PASSWORD)
+            .await
+            .expect("inject_admin_user");
+
+        assert!(
+            !k.domain_operations
+                .backend_tx_manager
+                .is_structure_seed_pending()
+                .await
+                .expect("read pending"),
+            "a seed obligation was armed on a server with nothing to seed. It would never be \
+             discharged, and a later deploy that added a structure config would inject the \
+             defaults into this by-then-established workspace."
         );
     }
 
--- a/citadel-workspace-server-kernel/src/kernel/async_kernel.rs
+++ b/citadel-workspace-server-kernel/src/kernel/async_kernel.rs
@@ -527,6 +527,20 @@ impl<R: Ratchet + Send + Sync + 'static> AsyncWorkspaceServerKernel<R> {
         // leaves a legitimately empty tree, and a contents-based check would then resurrect
         // the baked-in defaults on the next restart - re-applying config over structure the
         // users deliberately removed, which is exactly the contract above.
+        // CONCURRENCY: this is a check-then-act, and it is safe because the workspace server is
+        // a SINGLE-WRITER process. `on_start` is a once-per-process NetKernel lifecycle hook,
+        // and the server is one container owning one TCP listener - a second instance against
+        // the same backend cannot even bind. There is deliberately no compare-and-set here:
+        // the backend surface is a plain get/put over the SDK's `BackendHandler` (no CAS
+        // primitive exists to build on), and every other persisted sentinel in this kernel -
+        // `KEY_MIGRATION_DONE` in `migrate_if_needed`, and the schema-version get/set later in
+        // this same `on_start` - has exactly the same shape.
+        //
+        // If multi-process servers over one backend are ever introduced, this marker is NOT the
+        // thing to fix first: every node mutation is a load-mutate-save of the single
+        // `citadel_workspace.nodes` key with no cross-process locking, so concurrent writers
+        // would lose whole subtrees regardless of how this guard is written. That would need a
+        // backend-level locking story, not a special case here.
         if self
             .domain_operations
             .backend_tx_manager
@@ -545,6 +559,12 @@ impl<R: Ratchet + Send + Sync + 'static> AsyncWorkspaceServerKernel<R> {
         // duplicate its entire tree on the next boot - reintroducing the very bug this guard
         // exists to prevent. Any node parented to the workspace root proves a tree is already
         // established (config-seeded or user-created): record the marker and leave it alone.
+        //
+        // This path is crash-safe by construction: it writes ONLY the marker, and only when a
+        // tree already exists. It never writes nodes, so there is no partial state to unwind.
+        // A crash between reading `nodes` and writing the marker simply leaves the store as it
+        // was; the next boot re-reads the same tree and writes the marker then. Re-running it is
+        // a no-op, so a stale snapshot cannot cause incorrect behaviour - at worst it repeats.
         if nodes
             .values()
             .any(|node| node.parent_id.as_deref() == Some(crate::WORKSPACE_ROOT_ID))
--- a/citadel-workspace-server-kernel/src/kernel/async_kernel.rs
+++ b/citadel-workspace-server-kernel/src/kernel/async_kernel.rs
@@ -297,6 +297,26 @@ impl<R: Ratchet + Send + Sync + 'static> AsyncWorkspaceServerKernel<R> {
         if !workspace_exists {
             info!(target: "citadel", "Creating root workspace with no owner (first user with master password becomes admin)");
 
+            // Record the seed obligation BEFORE any of the durable writes below.
+            //
+            // ORDER IS THE WHOLE POINT. These are independent backend writes with no transaction
+            // around them, so whichever runs first defines the failure mode of the gap between
+            // them, and the two orderings are NOT symmetric:
+            //
+            //   * marker, then workspace - if we die in between, the next boot sees no workspace,
+            //     creates it, and (the marker being already set) seeds it. RECOVERABLE.
+            //   * workspace, then marker - if we die in between, the next boot sees a workspace
+            //     with NEITHER marker. That is indistinguishable from an established pre-marker
+            //     store, so the back-fill in `initialize_workspace_structure` stamps it "seeded"
+            //     and the workspace is left with no offices FOREVER. UNRECOVERABLE.
+            //
+            // So the marker goes first. A pending marker with no workspace is a harmless, self-
+            // healing state; a workspace with no marker is a permanent one.
+            self.domain_operations
+                .backend_tx_manager
+                .mark_structure_seed_pending()
+                .await?;
+
             // Debug: Check current storage mode
             if self.domain_operations.backend_tx_manager.is_test_mode() {
                 warn!(target: "citadel", "Creating workspace in test storage mode!");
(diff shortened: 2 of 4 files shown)


Tjemmmic:

> **Resolution (706d46f):** 2 addressed. 

### Summary

The PR adds a seed-once guard to prevent workspace structure duplication across server restarts against persistent backends, fixing a real data corruption bug with comprehensive regression tests. While the core fix correctly prevents the unbounded tree growth described in the PR, the freshness check has a logical loophole: it infers emptiness from the absence of root children, meaning a workspace where users have intentionally deleted all root offices will be misidentified as "fresh" and re-seeded with defaults on the next restart.

### Issues Found
**2 total** — 0 P1 (blocking) · 1 P2 (should fix) · 1 P3 (nice to have)

#### ~~⚠️ P2 — Seed guard re-injects defaults if users delete all root children~~ ✅ **Addressed**

> **Note:** Correct, and the original justification did not hold up: `save_nodes` is a single atomic key write, so the partial-seed failure the contents-based check was meant to self-heal cannot really occur, while the delete-all resurrection is real. Replaced it with a durable `citadel_workspace.structure_seeded` marker (mirroring the existing `KEY_MIGRATION_DONE` / schema-version pattern), plus the requested seed -> delete-all -> restart regression test. Note the marker alone is not sufficient: a store written before the marker existed is populated but unmarked, so a marker-only check would treat it as fresh and duplicate its tree. The guard therefore back-fills the marker when an existing root-parented tree is found, covered by its own test. (706d46f)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:528-536`
- **Problem:** The guard treats the workspace as fresh whenever no node has `parent_id == WORKSPACE_ROOT_ID`. If users intentionally delete all root offices via existing mutation paths, the workspace becomes empty of root children. The next server boot will run this function, misidentify the empty workspace as "fresh", and seed the baked-in offices and rooms again. This violates the stated contract that the configuration is only the initial state and must not be re-applied over user-owned structure.
- **Fix:** Persist an explicit durable seed marker (e.g., a backend transaction key like `citadel_workspace.structure_seeded`) instead of inferring freshness from current tree contents. Skip seeding when the marker is set, and set it after the first seeding attempt succeeds. Additionally, add a regression test for the sequence: seed -> delete all root children -> restart.

---

#### ~~ℹ️ P3 — Test relies on unverified hardcoded node count~~ ✅ **Addressed**

> **Note:** Replaced `after_first.len() == 2` with assertions on node NAMES. This keeps the non-vacuity guarantee (if seeding silently produced nothing, the idempotency check would compare 0 to 0 and pass while testing nothing) without breaking if the storage layer ever persists an extra structural node. (706d46f)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:1423-1428`
- **Problem:** The test `second_seed_does_not_duplicate_the_tree` asserts `after_first.len() == 2` based on the assumption that exactly two nodes (one office + one room) are created and no wrapper or root nodes are persisted by `save_nodes`. If the underlying schema persists the root node itself or creates structural wrappers, this assertion may fail for reasons unrelated to the regression being tested, masking the actual idempotency check.
- **Fix:** Filter out the workspace root or any structural wrapper nodes before asserting the count, or loosen the exact count assertion and focus strictly on verifying that the node keys and count remain identical across multiple seed attempts.

---

### ✅ APPROVE

The change is well-documented and successfully fixes a severe data corruption bug where persistent backends accumulated duplicate workspace trees on every restart. The core logic correctly prevents duplication when the workspace is already populated. The P2 issue regarding the "delete-all" edge case represents an incomplete fulfillment of the seed-once contract rather than a strict regression, and does not block merging this significant stability improvement. The P3 test quality issue is minor. We recommend addressing the delete-all edge case in a follow-up.

### Quick Reference
- **P2**: Seed guard re-injects defaults if users delete all root children
- **P3**: Test relies on unverified hardcoded node count

---

Tjemmmic:

> **Resolution (c75e151):** 2 not an issue. 

### Summary

This PR introduces a durable seed-once marker to prevent workspace structure duplication across server restarts, successfully resolving the original infinite duplication bug. It includes a back-fill check for pre-marker stores and comprehensive regression tests. However, the core check/seed/mark sequence is non-atomic, leaving the guarantee broken under concurrent startup or process crashes.

### Issues Found

#### ~~⚠️ P2 — Seed marker check and save sequence is non-atomic, duplicating structure during concurrent startup~~ ❌ **Not an issue**

> **Note:** The check-then-act is real, but the scenario is not. The workspace server is a single-writer process: `on_start` is a once-per-process NetKernel hook, and the server is one container owning one TCP listener - a second instance against the same backend could not even bind :12349. There is also no primitive to build the suggested CAS on: the backend surface is a plain get/put over the SDK's `BackendHandler`, so `try_mark_structure_seed_in_progress_or_seeded()` would be an SDK-level change, not a kernel one. Every other persisted sentinel here has exactly this shape already - `KEY_MIGRATION_DONE` in `migrate_if_needed`, and the schema-version get/set later in this same `on_start`. And if multi-process servers over one backend were ever introduced, this marker would not be the thing to fix first: every node mutation is a load-mutate-save of the single `citadel_workspace.nodes` key with no cross-process locking, so concurrent writers would lose whole subtrees regardless of how this guard is written. That needs a backend locking story, not a special case here. The single-writer assumption is now documented at the call site. (c75e151)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:527-756`
- **Problem:** The `initialize_workspace_structure` function checks the marker, builds the tree with new UUIDs, saves it, and only then writes the marker without any locks or compare-and-set operations. If two server instances start against the same fresh persistent workspace concurrently, both can observe `false` and no root children, leading both to seed independently and duplicate the tree.
- **Fix:** Make the seed decision and marker write atomic at the backend level. Add a backend operation such as `try_mark_structure_seed_in_progress_or_seeded()` that succeeds for exactly one caller, or run the marker creation and node write within a transaction/lock supported by the persistent backend.

---

#### ~~⚠️ P2 — Back-fill marker write is non-atomic with seed save, risking partial state on crash~~ ❌ **Not an issue**

> **Note:** The back-fill path writes ONLY the marker, and only when a tree already exists - it never writes nodes, so there is no partial state to unwind and nothing to roll back. A crash between reading `nodes` and writing the marker leaves the store exactly as it was; the next boot re-reads the same tree and writes the marker then. The review identifies this itself ("forces a redundant back-fill on the next boot"), and a redundant back-fill is a no-op. Taking the review's second suggested remedy: the idempotency and crash-safety of this path are now explicitly documented. (c75e151)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:558-562`
- **Problem:** The back-fill `mark_structure_seeded()` call operates on a snapshot of nodes from `get_all_nodes()`. If the process crashes after the legacy tree is evaluated but before the back-fill marker is fully written, the workspace could be left in a state where the marker exists but the tree is unexpectedly empty if rolled back, or forces a redundant back-fill on the next boot.
- **Fix:** Wrap the back-fill marker write in the same transaction context as the node read, or explicitly document that the back-fill path relies on the snapshot from `get_all_nodes()` being sufficiently fresh for idempotency.

---

> **Status:** Both concerns describe a multi-writer scenario the deployment does not have (single container, single TCP bind, once-per-process `on_start`), and the suggested CAS primitive does not exist in the backend abstraction. The single-writer assumption and the back-fill's crash-safety are now documented at the call site.

The PR effectively solves the duplication bug for standard single-process restarts and includes excellent documentation and tests. However, the implementation retains a non-atomic check-then-act sequence that preserves the original duplication bug under concurrent startup, which defeats the core guarantee of this PR. Additionally, the back-fill marker write is non-atomic with the node snapshot read. Resolving these concurrency and atomicity issues requires another pass before merge.

### Quick Reference
- **P2**: Seed marker check and save sequence is non-atomic, duplicating structure during concurrent startup
- **P2**: Back-fill marker write is non-atomic with seed save, risking partial state on crash

---

Tjemmmic:

> **Resolution (744ce9e):** 2 addressed. 

### Summary

The PR introduces a durable marker to prevent workspace structure duplication on restarts against persistent backends, along with backfill logic for pre-marker stores and regression tests. The fix is generally well-reasoned and documented. However, there is a blocking edge case in the upgrade path: pre-marker workspaces that have been legitimately emptied by an admin (zero nodes parented to the root) will fail the backfill check, fall through to the fresh-seed path, and resurrect the baked-in defaults.

### Issues Found
**2 total** — 1 P1 (blocking) · 1 P2 (should fix) · 0 P3 (nice to have)

#### ~~❗ P1 — Pre-marker empty workspaces are still re-seeded on upgrade~~ ✅ **Addressed**

> **Note:** Correct, and a real gap - thank you. The back-fill inferred "established" from node CONTENTS, but an established workspace whose offices were all deleted has no root children either, so it fell through to the fresh-seed path and resurrected the defaults. Fixed by taking the creation fact directly instead of inferring it: `inject_admin_user` now returns whether it CREATED the root workspace on this boot, and `initialize_workspace_structure` seeds only when that is true. An empty tree and a new workspace are now distinct facts. Added `emptied_legacy_workspace_is_not_reseeded_on_upgrade`, verified non-vacuous - against the previous contents-based logic it fails with "the baked-in defaults were resurrected into an established workspace that an admin had emptied (2 nodes reappeared)". The contents check is retained only as defence-in-depth, since it can add a skip but never cause a seed. (744ce9e)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:558-569`
- **Problem:** The backfill guard only recognizes legacy workspaces if at least one node has `parent_id == WORKSPACE_ROOT_ID`. A production workspace created before `KEY_STRUCTURE_SEEDED` existed can legitimately have zero offices because an admin deleted them all. On upgrade, it has no marker and no root-child nodes, so this code falls through to the fresh-seed path and recreates the baked-in defaults. This directly violates the stated contract that deleting every office must survive restart.
- **Fix:** Distinguish a genuinely new workspace from an existing pre-marker workspace using workspace creation state rather than node contents alone. For example, check if the root workspace already exists; if it does and the marker is absent, treat it as a legacy store by writing `KEY_STRUCTURE_SEEDED` and skipping the seed, even when the node map is empty.

---

#### ~~⚠️ P2 — Regression tests do not actually test the persistent backend scenario~~ ✅ **Addressed**

> **Note:** The gap is real, but the suggested unit test is not constructible: `backend_get`/`backend_save` route to the in-memory `test_storage` whenever no `NodeRemote` is set, and a unit test cannot supply one - which is why the existing `migration_tests` module states that in-memory "is the only mode reachable from a unit test" and defers real-backend behaviour to integration tests. So the coverage was added where the filesystem backend actually runs: a new step in the `production-build` job restarts the server against the SAME `server_data` volume and asserts the office/room creation counts did not grow. That is precisely the check that catches this bug class, and it also fails loudly if the first boot seeds nothing (so it cannot pass vacuously). This closes the exact CI hole that let the original bug ship - every other job runs the server in-memory, where every boot legitimately starts empty. (744ce9e)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:1428-1665`
- **Problem:** The PR description notes that dev and CI never caught the original bug because they run on the in-memory backend. While the new tests verify the guard logic and idempotency within a single kernel instance using the in-memory backend, they do not verify that `backend_save(KEY_STRUCTURE_SEEDED, &true)` actually persists the marker to a filesystem backend so it survives a process restart. If the persistence of the marker key fails on the real backend, these tests would still pass while the bug persists in production.
- **Fix:** Add at least one integration test that verifies a second kernel instance pointing at the same persistent (filesystem) backend reads the marker correctly and skips seeding.

---

> **Status:** The P1 was a genuine gap and is fixed at the root (creation fact, not node contents), with a non-vacuous regression test. The persistent-backend scenario is now covered by a real filesystem-backend restart check in CI.

The core fix is sound, but the PR still resurrects configured defaults for established legacy workspaces that happen to be empty during an upgrade. Because this PR's explicit goal is to prevent unwanted structure writes and respect user deletions across restarts, the P1 edge case must be resolved before merging to avoid repeating the exact data-integrity issue the PR aims to fix.

### Quick Reference
- **P1**: Pre-marker empty workspaces are still re-seeded on upgrade
- **P2**: Regression tests do not actually test the persistent backend scenario

---

Tjemmmic:

> **Resolution (9794db1):** 3 addressed. 

### Summary

The PR introduces a durable marker to ensure workspace structure is seeded only once per persistent backend, fixing a duplication bug on restarts. The core logic is well-documented and addresses concurrency and migration concerns, but the lack of integration testing for the new `workspace_is_new` boot-time wiring leaves a gap in the regression guard. Additionally, a brittle log-based CI test and a potential interrupted-first-boot edge case warrant follow-up.

### Issues Found
**3 total** — 0 P1 (blocking) · 3 P2 (should fix) · 0 P3 (nice to have)

#### ~~⚠️ P2 — Missing integration coverage for the new boot-time seed decision~~ ✅ **Addressed**

> **Note:** Agreed - hand-feeding the decision into the function under test left the real wiring unguarded. Every test now drives the actual boot sequence through a `boot()` helper that calls `inject_admin_user` then `initialize_workspace_structure`, exactly as `on_start` does, so a mistake in the workspace-exists detection breaks the tests rather than only production. Added `inject_admin_user_records_the_seed_obligation_exactly_once`, which pins the decision boundary directly: a virgin store owes nothing, creating the root workspace takes on the obligation, seeding discharges it, and a restart must not re-arm it. The legacy-store tests now build a genuine pre-marker workspace (root workspace present, neither marker set) rather than simulating one with a flag. (9794db1)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:L851-L861`
- **Problem:** The production behavior now depends on `inject_admin_user` returning the correct `workspace_is_new` value and `on_start` passing that value into `initialize_workspace_structure`. However, the added unit tests call `initialize_workspace_structure(..., true/false)` directly. A future or existing mistake in the `inject_admin_user` workspace-exists detection, or in the `on_start` wiring, would bypass the regression tests while breaking the re-seed skip logic in production.
- **Fix:** Add a focused test that exercises the real boot decision boundary instead of manually supplying the boolean. At minimum, test `inject_admin_user` on the same backend to verify it returns `true` on first initialization and `false` on the second. Feed those returned values into `initialize_workspace_structure` to assert the tree is created exactly once.

---

#### ~~⚠️ P2 — Interrupted first boot can mark a fresh workspace as seeded without creating the structure~~ ✅ **Addressed**

> **Note:** Real, and worse than described: it needs no crash at all. A transient failure of the node write aborts `on_start`, the container restarts, and boot 2 lands in exactly this state - root workspace present, no offices, no marker - so the back-fill would stamp it seeded and the workspace would be left with NO offices permanently. Fixed with a durable `structure_seed_pending` marker written by `inject_admin_user` at the moment it creates the root workspace and cleared only once the tree is durably written. An interrupted first boot therefore resumes instead of being abandoned. This also let the design get simpler: the marker subsumes the `workspace_is_new` parameter entirely, so that parameter is gone and the decision is driven purely by durable state. Regression test `interrupted_first_boot_is_resumed_not_abandoned`, verified non-vacuous - without the pending marker it fails with "an interrupted first boot was abandoned: the workspace was left with no offices (0 nodes)". (9794db1)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:L587-L596`
- **Problem:** The `!workspace_is_new` branch treats every store without `KEY_STRUCTURE_SEEDED` as an established legacy workspace, writes the seeded marker, and returns. If the first boot creates the root workspace but crashes or fails before `initialize_workspace_structure` completes, the next boot will report `workspace_is_new == false`. It will hit this branch with an empty node map, mark the structure as seeded, and permanently skip creating the configured offices and rooms.
- **Fix:** Persist a separate durable signal for pending structure seeding, or only use the legacy backfill path for stores definitively lacking the new-workspace creation signal. Add a regression test covering a crash immediately after root workspace creation but before structure seeding.

---

#### ~~⚠️ P2 — CI persistence test greps container log text instead of checking workspace state~~ ✅ **Addressed**

> **Note:** Fair - comparing cumulative counts before and after coupled the result to docker's log buffering and rotation, and if early lines were dropped the "after" count could come out LOWER and misfire. The check now captures a timestamp and inspects only `--since` that window, asking the precise question - "did THIS boot create any office or room nodes?" - to which the answer must be zero. It also now requires POSITIVE confirmation that the guard fired (the skip line must appear), so a server that never reached the seeding code cannot pass by simply producing no output, and it still fails loudly if the first boot seeded nothing (which would make the whole check vacuous). Querying workspace state directly would need a protocol client in that step; the log signal, correctly scoped, is the available evidence. (9794db1)

- **File:** `.github/workflows/validate.yml:510-L546`
- **Problem:** The integration test counts 'Creating office node' / 'Creating room node' log lines from `$compose logs server` to detect re-seeding. Depending on `docker compose logs` to accurately buffer, persist, and truncate log output for exact counts is brittle and couples the test's pass/fail state to log string matching rather than actual workspace state.
- **Fix:** If an API endpoint or DB query is available to enumerate offices, query that instead of grep-ing logs. If logs are the only signal, constrain the grep to the post-restart timeframe using `docker compose logs --since` or by capturing a timestamp before restarting.

---

> **Status:** All three P2s fixed. The interrupted-first-boot hole is closed with a durable pending marker (which also removed the `workspace_is_new` parameter), the tests now drive the real `inject_admin_user` -> `initialize_workspace_structure` boot path, and the CI persistence check uses a scoped post-restart window plus positive confirmation that the guard fired.

The core Rust logic and unit tests are thorough and well-reasoned, directly addressing the restart duplication bug with safe back-fill mechanics. However, per the decision rules, a pull request with 3 or more P2 issues requires changes. The missing integration test for the newly wired boot path, the brittle CI log parsing, and the potential interrupted-first-boot edge case leave significant gaps in the regression coverage and reliability of this critical startup logic.

### Quick Reference
- **P2**: Missing integration coverage for the new boot-time seed decision
- **P2**: Interrupted first boot can mark a fresh workspace as seeded without creating the structure
- **P2**: CI persistence test greps container log text instead of checking workspace state

---

Tjemmmic:

> **Resolution (99c3936):** 1 addressed. 

### Summary

The PR adds durable pending/seeded markers to prevent workspace structure duplication on restart, with a robust regression suite and CI check. However, reviewers B and C correctly identified a critical ordering flaw: the pending marker is written AFTER the root workspace, domain, and password are persisted, leaving a failure window that can permanently classify a fresh workspace as an established pre-marker store and skip seeding forever. This directly breaks the interrupted-first-boot recovery contract.

### Issues Found
**1 total** — 1 P1 (blocking) · 0 P2 (should fix) · 0 P3 (nice to have)

#### ~~❗ P1 — Pending marker written after durable workspace creation, permanently breaking recovery~~ ✅ **Addressed**

> **Note:** Correct, and a sharp catch - the fix I shipped last round still had the failure window it was meant to close, just moved. The marker write and the workspace writes are independent (no transaction spans them), and the two orderings are NOT symmetric: marker-then-workspace leaves a marker with no workspace, which self-heals (the next boot creates the workspace and, the marker being set, seeds it); workspace-then-marker leaves a workspace with NEITHER marker, which is indistinguishable from an established pre-marker store, so the back-fill stamps it seeded and the workspace is left with no offices forever. `mark_structure_seed_pending()` now runs FIRST, before `insert_workspace` / `insert_domain` / `set_workspace_password`, with the asymmetry documented at the call site so the ordering is not silently 'tidied' later. Added `seed_obligation_is_recorded_before_the_workspace_is_persisted`, which reproduces the interrupted state directly (marker set, workspace never persisted) and asserts the next boot still produces a fully seeded workspace. 150 tests pass. (99c3936)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:340-350`
- **Problem:** `inject_admin_user` writes the `structure_seed_pending` marker only after calling `insert_workspace`, `insert_domain`, and `set_workspace_password`. If the process crashes or the marker write fails after those earlier writes succeed, the next boot sees an existing workspace with neither `structure_seeded` nor `structure_seed_pending`. `initialize_workspace_structure` then takes the pre-marker backfill path, marks the workspace seeded, and skips creating offices/rooms forever. This leaves a genuinely fresh workspace permanently unseeded—the exact interrupted-first-boot failure the PR claims to fix.
- **Fix:** Move `mark_structure_seed_pending().await?` to the beginning of the `if workspace.is_none()` creation branch, before any durable root-workspace/domain/password writes. A pending marker without a workspace is recoverable (next boot retries creation), but a workspace without the pending marker is not. Add a regression test simulating a workspace persisted without either seed marker to verify it doesn't get backfilled as seeded without structure.

---

> **Status:** The ordering flaw is fixed: the seed obligation is now recorded before any durable workspace write, so the gap between them is self-healing rather than permanent, and a regression test pins it.

### Quick Reference
- **P1**: Pending marker written after durable workspace creation, permanently breaking recovery

---

Tjemmmic:

> **Resolution:** 3 addressed. 

### Summary

The PR introduces durable seed-once markers to prevent workspace structure duplication on persistent backends, with sound core logic and excellent test coverage for interrupted first boots and deletion scenarios. However, three distinct behavior bugs remain: the defensive check leaves the pending marker dangling, the pending marker is armed even when no structure config is provided, and the cleanup operation can fail an otherwise successful boot.

### Issues Found
**3 total** — 0 P1 (blocking) · 2 P2 (should fix) · 1 P3 (nice to have)

#### ~~⚠️ P2 — Defensive check marks workspace as seeded without clearing the pending marker~~ ✅ **Addressed**

> **Note:** Right - that branch half-completed the transition and left a dangling `pending`. It now marks the workspace seeded AND discharges the obligation, so no stale flag is left for a future reader to puzzle over. (Note the premise about partially-written nodes cannot actually occur: `save_nodes` is a single atomic key write, so the tree is either fully written or not at all. The dangling marker was real regardless.) (21fc5f8)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:620-628`
- **Problem:** In the 'defence in depth' branch, when `nodes` already has root children but `seed_pending` is true, the code calls `mark_structure_seeded()` and returns. If a first boot was interrupted AFTER writing some (but not all) nodes, the workspace is marked as fully seeded while the pending marker remains set. This creates an inconsistent state where the workspace tree is incomplete and the pending marker is left permanently dangling.
- **Fix:** Also clear the pending marker for consistency in the defence-in-depth branch: add `self.domain_operations.backend_tx_manager.clear_structure_seed_pending().await?;` before returning.

---

#### ~~⚠️ P2 — Pending seed marker is armed even when no structure config will run~~ ✅ **Addressed**

> **Note:** A genuine bug, and a nasty one. `on_start` only calls `initialize_workspace_structure` when a structure is configured, so a server first booted WITHOUT one would arm an obligation that nothing would ever discharge - it would sit there while the workspace ran and accumulated real content, and the first deploy that ADDED a structure config would find `pending == true` and inject the baked-in defaults into that by-then-established workspace. Exactly what this guard exists to prevent. `inject_admin_user` now arms the obligation only when `self.workspace_structure.is_some()`: the debt is only taken on if something is going to pay it. Regression test `no_structure_config_means_no_seed_obligation`, verified non-vacuous - arming it unconditionally makes it fail. The test kernel is also now wired the way a real server is (with a structure configured), so this path is exercised rather than assumed. (21fc5f8)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:297-316`
- **Problem:** `inject_admin_user` records `structure_seed_pending` whenever it creates the root workspace, but `on_start` only calls `initialize_workspace_structure` when `self.workspace_structure` is `Some`. If the server first boots with no structure config, the workspace is created and runs as an established empty workspace while the pending marker remains true. A later deploy that adds a config will see `seed_pending == true` and inject the baked-in defaults into that existing workspace, violating the PR's contract.
- **Fix:** Only arm the pending marker when the root workspace is being created for a boot that actually has a structure config to seed. Pass a `seed_structure_pending: bool` flag from `on_start` based on `self.workspace_structure.is_some()` into the workspace-creation path, and only write the marker when that flag is true.

---

#### ~~ℹ️ P3 — Nonessential pending-marker cleanup can fail a successful first boot~~ ✅ **Addressed**

> **Note:** Agreed - the tree is durably written and the workspace durably marked seeded by that point, so failing the boot over a cosmetic write would turn a completed seed into a crash loop, and the next boot would short-circuit on `seeded` anyway. The clear is now best-effort: a failure logs a warning and the boot proceeds. (21fc5f8)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:842-849`
- **Problem:** After `save_nodes` and `mark_structure_seeded` succeed, `clear_structure_seed_pending().await?` is treated as fatal even though the comment says clearing pending is hygiene and `seeded` already takes precedence. A transient backend error on this cleanup write makes `initialize_workspace_structure` return an error after the workspace tree is already durably seeded and marked complete, causing an unnecessary startup failure.
- **Fix:** Do not propagate cleanup failure after the seeded marker is written. Log a warning and return `Ok(())`, making the clear operation best-effort so the successful seed is not reported as a failed boot.

---

> **Status:** All three fixed. The obligation is now armed only when a structure is configured to discharge it, the defence-in-depth branch completes the transition instead of leaving a dangling marker, and clearing the marker is best-effort so it cannot fail an otherwise successful boot. 151 tests pass.

Three P2 issues were identified across the independent reviews, each representing a distinct behavior bug in the new persistence logic. The most severe is the arming of the pending marker without a structure config (which can resurrect defaults into an established workspace) and the defensive check's inconsistent state. These contradict the PR's own stated contracts around resumability and non-resurrection, requiring another pass before merge.

### Quick Reference
- **P2**: Defensive check marks workspace as seeded without clearing the pending marker
- **P2**: Pending seed marker is armed even when no structure config will run
- **P3**: Nonessential pending-marker cleanup can fail a successful first boot

---

Tjemmmic:

> **Resolution:** 2 addressed. 

### Summary

The PR adds durable 'seeded' and 'seed pending' markers to prevent workspace structures from duplicating on server restarts against persistent backends. The implementation is logically sound with excellent handling of crash-recovery edge cases and great regression test coverage. The identified issues are minor error-handling and hygiene inconsistencies that do not block the merge.

### Issues Found
**2 total** — 0 P1 (blocking) · 0 P2 (should fix) · 2 P3 (nice to have)

#### ~~ℹ️ P3 — Recovery path treats pending-marker cleanup as fatal after the workspace is already marked seeded~~ ✅ **Addressed**

> **Note:** Correct - the recovery branch still propagated the cleanup error with `?` while the normal seeding path had already been made best-effort, so the same cosmetic write was fatal in one place and optional in another. That is exactly the kind of drift that bites later. All three discharge sites now go through a single `discharge_seed_obligation()` helper that logs and swallows a failure, so they cannot diverge again. (f2a05de)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:640-646`
- **Problem:** In the `seed_pending && existing tree` recovery branch, the code writes `mark_structure_seeded()` and then propagates any error from `clear_structure_seed_pending()` with `?`. At that point, the authoritative `seeded` marker has already been written, and the workspace is safely recoverable. Failing the current boot over a cosmetic cleanup write is inconsistent with the normal seeding path below, which correctly treats the same operation as best-effort.
- **Fix:** Handle `clear_structure_seed_pending()` the same way as the normal successful seeding path: log a warning on error and return `Ok(())` after `mark_structure_seeded()` succeeds.

---

#### ~~ℹ️ P3 — Early return for already-seeded workspaces leaves a dangling 'pending' marker~~ ✅ **Addressed**

> **Note:** Agreed. A leftover `pending` can only exist if an earlier discharge failed (it is best-effort), so the already-seeded path now opportunistically clears it. Both the read AND the write there are best-effort - `seeded` is authoritative, so nothing on that path may fail a boot over a cosmetic flag. The store is now self-healing rather than carrying a contradictory pair of markers indefinitely. (f2a05de)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:556-565`
- **Problem:** If `is_structure_seeded()` returns true, the function logs and returns `Ok(())` early. While `seeded` takes precedence on all read paths, leaving a stale `pending == true` marker behind introduces an unclean state that could confuse future logic or make state debugging harder.
- **Fix:** Defensively clear the `pending` marker (handling any error as best-effort) before returning `Ok(())` when the workspace is already marked as seeded.

---

### ✅ APPROVE

The PR correctly resolves the structure duplication bug with a robust, well-reasoned durable marker approach. All reviewers approved the PR, and the identified issues are minor P3 inconsistencies regarding best-effort cleanup of the 'pending' marker. They do not affect the correctness of the core logic or crash recovery guarantees, so they can be safely addressed in a follow-up cleanup.

### Quick Reference
- **P3**: Recovery path treats pending-marker cleanup as fatal after the workspace is already marked seeded
- **P3**: Early return for already-seeded workspaces leaves a dangling 'pending' marker

---

Tjemmmic:

> **Resolution (972db5a):** 2 not an issue (1 test added). 

### Summary

The PR adds durable seed markers to make workspace structure initialization idempotent, preventing duplicate tree creation on restarts. Reviewer A identifies a P1 issue where the defense-in-depth check contradicts the PR's stated intent for resuming interrupted first boots, and a P2 issue about CI log strings that may not exist in the codebase. Both issues are supported by the diff and warrant fixes before merge.

### Issues Found
**2 total** — 1 P1 (blocking) · 1 P2 (should fix) · 0 P3 (nice to have)

#### ~~❗ P1 — Defense-in-depth check freezes partial seeds instead of completing them~~ ❌ **Not an issue**

> **Note:** The premise does not hold: a partial tree cannot exist. `save_nodes` serialises the ENTIRE node map into one JSON blob and writes it under a single key (`citadel_workspace.nodes` - see `transaction/mod.rs`), so either the whole map lands or none of it does. There is no "first office written, the rest not" state to repair, which is why the branch is safe.

More importantly, the suggested fix would REINTRODUCE the bug this PR exists to fix. Firing the check only when the pending marker is ABSENT makes it dead code - that case already returns earlier via the pre-marker back-fill. So the suggestion reduces to deleting the branch, and deleting it sends this exact state (seed owed + tree present) down the seeding path, depositing a second copy of the whole tree on top of the first. Verified, not argued: disabling the branch makes the new test fail with "the existing tree was seeded over rather than recorded: 3 nodes now exist" - the tree grew from 1 node to 3.

The review is right about one thing, and it is the valuable part: this branch had no test. `tree_written_but_seed_unrecorded_is_finalised_not_duplicated` now covers it - it reproduces a boot that wrote the nodes and died before recording the seed, and asserts the next boot finalises that transition (marks seeded, discharges the obligation) and leaves the tree byte-for-byte intact. The comment now also states the all-or-nothing property explicitly, so the branch's safety is not left to be re-derived. (972db5a)

- **File:** `citadel-workspace-server-kernel/src/kernel/async_kernel.rs:668-686`
- **Problem:** The comment at L648-L655 states that the pending marker makes an interrupted first boot resumable: 'A crash ... between creating the root workspace and writing the tree would otherwise leave a store that looks exactly like an established pre-marker workspace ... and would be left with NO offices, permanently.' However, the defense-in-depth check at L668-L686 short-circuits ANY resume when ANY parent_id==ROOT node exists: it marks seeded and returns WITHOUT seeding. Consider a partial save_nodes that wrote the first office before failing—the tree is incomplete, the seed is owed, yet this guard stamps it 'seeded' and abandons the rest. The comment claims the check is 'safe to keep' because it 'can only ever ADD a skip', but adding a skip to a partial seed is precisely the unrecoverable state the pending marker was designed to prevent. The interrupted_first_boot_is_resumed_not_abandoned test does not cover this case because it never writes any nodes before the simulated interruption.
- **Fix:** Either remove the defense-in-depth check, or change it to only fire when the pending marker is ABSENT (genuine pre-marker back-fill). When seed_pending is true, any partial tree must be completed, not frozen. Alternatively, make save_nodes all-or-nothing so partial trees are impossible, which would make the existing check safe.

---

#### ~~⚠️ P2 — CI check greps log strings that may not exist in the codebase~~ ❌ **Not an issue**

> **Note:** All three strings exist. `"Creating office node '{}' (id: ...)"` is emitted at async_kernel.rs:785 and `"Creating room node '{}' in office '{}'"` at :861 - both inside the seeding loop, which the diff view truncated. The skip line is `"Initial workspace structure already seeded; skipping"` at :614, and the CI greps for `already seeded; skipping`, which matches it as a substring (grep is substring matching, not exact-line). These are also not theoretical: the same log lines were observed verbatim from the real server container during the filesystem-backend restart runs that reproduced and then verified the fix. The vacuous-pass concern is separately guarded - the step fails if the FIRST boot logs no office creations at all, so it cannot pass by matching nothing. (no change needed)

- **File:** `.github/workflows/validate.yml:527,553`
- **Problem:** The persistence test asserts on log substrings 'Creating office node' and 'Creating room node'. However, the diff does not show these exact strings anywhere in the seeding code path—the only related log message visible is 'Workspace structure initialization complete' at L896. The 'already seeded; skipping' string is also not exact—the actual log message is 'Initial workspace structure already seeded; skipping' at L599. If the grep patterns don't match the actual log output, the vacuous-guard will fail the step, or if no offices are logged the check passes vacuously once the guard is removed.
- **Fix:** Verify the exact log strings emitted by the office/room creation path in async_kernel.rs (lines after save_nodes at L879, not shown in this diff). Update the grep patterns to match actual strings, or add structured log fields (e.g., action = "seed_office") and grep on those to decouple from message wording.

---

> **Status:** Neither holds. The P1 rests on a partial `save_nodes`, which cannot occur (one key, one blob, all-or-nothing), and its suggested fix would delete a branch whose removal demonstrably duplicates the tree (1 node -> 3 in the new test). The P2's log strings all exist and were observed verbatim from a real server container. The one fair point - that the branch was untested - is now covered by a regression test.

The P1 issue is well-founded: the defense-in-depth check directly contradicts the PR's own stated contract for interrupted first boots. When seed_pending is true and a partial tree exists (e.g., save_nodes wrote some nodes before failing), the check marks it 'seeded' and abandons the incomplete seed—the exact unrecoverable state the pending marker was built to prevent. The fix is straightforward: only apply the defense-in-depth skip when the pending marker is absent. The P2 CI issue is also valid—the grep patterns reference log strings that may not match actual output, risking either false failures or vacuous passes. Two approving reviewers missed these issues, but the P1 correctness concern regarding partial tree handling is substantive and the fix is clear.

### Quick Reference
- **P1**: Defense-in-depth check freezes partial seeds instead of completing them
- **P2**: CI check greps log strings that may not exist in the codebase

---

Tjemmmic:

> **Resolution (972db5a):** No issues raised; nothing to disposition. 

### Summary

The PR introduces durable seed markers to ensure the configured workspace structure is only applied to genuinely fresh workspaces, preventing duplication on restart. It also adds robust recovery handling for interrupted first boots and includes extensive regression tests and a CI check against a real filesystem backend. The implementation is thoughtful and the accompanying tests are rigorous.

### No Issues Found

### ✅ APPROVE

All three independent reviewers agreed on an APPROVE, and the diff confirms a sound, well-documented implementation. The PR correctly uses durable markers instead of content-based checks, carefully orders backend writes for crash recovery, handles upgrade edge cases, and provides comprehensive regression coverage for the main failure modes. No P1 or P2 issues were identified.

---

PR #60, opened 2026-07-12
feat(deploy): publish slim images to GHCR and pull instead of building on the host

Tjemmmic:
## Overview

Deploys currently build images **on the production host** (`deploy.sh` runs `docker compose build`). That has four separate problems:

1. **The images are enormous.** The Rust Dockerfiles are single-stage (`FROM rust:1.92.0`, full stop), so the final images ship the entire Rust toolchain, the full source tree and `target/` to production: **server 5.65 GB, internal-service 4.43 GB**. The server binary is **29 MB**.
2. **Every deploy is a full Rust release build** on the box — many minutes of compile for what should be a file copy.
3. **No rollback.** The previous image is simply gone once rebuilt.
4. **Every deploy depends on the host's Docker build networking.** On the current deployment host it is broken: k3s/kube-router owns the iptables `FORWARD` chain (policy `DROP`), and Docker's bridge ACCEPT rule sits at position 13,274 having passed **zero packets ever**. The default bridge therefore has **no outbound connectivity at all** — not just DNS — so any `apt-get`/`npm` in a Dockerfile fails. Building on the host is hostage to that.

## Fix

Build images in CI, publish them to GHCR, and have the host **pull** instead of build. The host stops building entirely, which makes problem 4 irrelevant rather than working around it.

### 1. Slim multi-stage runtime images

Both Rust Dockerfiles gain a `production` target: a `rust:1.92.0` builder stage, then a `debian:trixie-slim` runtime stage carrying only the binary, its runtime assets and its shared libraries. The dependency list is derived from `ldd` against the built binaries (`libssl3`, `libzstd1`, `zlib1g` on top of what the slim base already has), plus `netcat-openbsd` — which is **not** optional, because the compose healthchecks probe the listener with `nc -z` and would otherwise never pass.

| Image | Before | After |
|---|---|---|
| workspace-server | 5.65 GB | **142 MB** |
| internal-service | 4.43 GB | **145 MB** |

This mirrors a pattern already in the repo — `docker/ui/Dockerfile` is already `dev` → `build` → `production`.

**Stage order is load-bearing.** `docker build` with no `--target` builds the *last* stage, and `docker-compose.yml` (dev / tilt / CI integration tests) pins **no target** for these two services, unlike `ui` which pins `target: dev`. So `dev` is deliberately kept as the **last** stage and remains equivalent to the previous single-stage image; production opts in explicitly with `target: production`. Putting `production` last would have silently handed every dev and CI build a toolchain-less image.

Also verified: despite the `rfd` dependency (native file dialogs), the internal-service binary has **zero GTK/GDK linkage**, so no desktop stack has to be dragged into its runtime.

### 2. Publish workflow

`.github/workflows/publish-images.yml` builds and pushes on merge to `master` (plus `workflow_dispatch`), tagging each image with both `latest` and an immutable `sha-<short>`.

### 3. Host pulls instead of builds

`docker-compose.production.yml` gains `image: ghcr.io/...:${IMAGE_TAG:-latest}` and `target: production`; `deploy.sh` pulls rather than builds. Rollback becomes:

```
IMAGE_TAG=sha-abc1234 ./deploy.sh --no-pull
```

`WORKSPACE_BIND_ADDR` is now overridable (still defaulting to loopback). The loopback default is correct for the co-located-cloudflared topology, but the local-agent topology needs the opposite: each user runs their own agent and dials the server directly, so it must bind `0.0.0.0`. The Citadel protocol is E2E encrypted, so a public bind is by design in that mode.

The now-orphaned `target_cache` volume is dropped from the production compose — the runtime images contain no cargo, no source and no `target/`, so there is nothing left for it to cache. The dev compose keeps its own.

## Why `ui` is not published here

The production UI image bakes `VITE_WS_URL` at build time and its nginx CSP is `connect-src 'self'` — so a UI served on `:8080` cannot reach an agent on `:12345`. Publishing it today would ship an artifact that **cannot work** for the local-agent deployment. It joins the workflow once the same-origin `/ws` reverse proxy (and the matching frontend change to derive the URL from `window.location`) lands. `deploy.sh` still builds `ui` locally in the meantime.

## Verification (on the real deployment host)

- Both `production` targets build and were **functionally tested**, not just sized: the slim server boots, engages the filesystem backend, seeds the workspace from `documents/`, listens, and passes an actual `nc -z` probe with **zero errors/panics**. The slim agent builds, runs and listens.
- The **default** target of both Dockerfiles still contains `cargo` — confirming the dev/Tilt/CI flow is unaffected.
- `docker compose config` validates; `IMAGE_TAG=sha-abc1234` and `WORKSPACE_BIND_ADDR=0.0.0.0:12349` both verified to take effect.
- `bash -n deploy.sh` passes.

## One-time manual step after merge

GHCR packages are created **private**. After the first successful run, set each package's visibility to Public in its GitHub settings, otherwise every developer must `docker login ghcr.io` before pulling. The repositories are already public; nothing in these images is secret.

## Follow-ups (not in this PR)

- The host's Docker bridge has no egress (k3s/kube-router `FORWARD` conflict). It needs root to fix and breaks *all* bridge Docker on that machine, not just this project.
- `update-avarok-server.sh` is superseded by this pipeline and should be retired.

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -168,6 +168,12 @@ jobs:
         run: |
           set -euo pipefail
           name="smoke-${{ matrix.image }}"
+          # Clean up on EVERY exit path, not just success. On a GitHub-hosted runner a leaked
+          # container dies with the VM, but this repository also has a SELF-HOSTED runner, where
+          # a container left behind by a failed run keeps its ports bound and poisons the next
+          # run. (A five-month-old leaked compose stack squatting on :12345/:12349 is exactly
+          # how the deployment host ended up blocked.)
+          trap 'docker rm -f "$name" >/dev/null 2>&1 || true' EXIT
           docker run -d --name "$name" ${{ matrix.smoke_env }} "$IMAGE:$SHA_TAG"
 
           ok=0
@@ -190,7 +196,6 @@ jobs:
             exit 1
           fi
           echo "${{ matrix.image }} is listening on ${{ matrix.smoke_port }}; image is good."
-          docker rm -f "$name" >/dev/null 2>&1 || true
 
       - name: Push immutable SHA tag
         # Only the immutable tag is pushed here. `latest` is advanced by the
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -7,13 +7,30 @@
 # `docker compose build`). That is slow (a full Rust release build per deploy),
 # it requires the source tree and a toolchain on the production box, it offers
 # no rollback (the previous image is simply gone), and it makes every deploy
-# hostage to that host's Docker networking. Building here instead and having the
-# host `docker compose pull` fixes all four at once.
+# hostage to that host's Docker build networking. Building here instead and
+# having the host `docker compose pull` fixes all four at once.
 #
-# Images are tagged with BOTH:
-#   * `sha-<short-sha>` - immutable. This is what makes rollback possible:
-#     `IMAGE_TAG=sha-abc1234 ./deploy.sh` redeploys an exact prior build.
-#   * `latest`          - moving pointer to the newest master build.
+# TAGGING AND THE `latest` CONTRACT
+#
+# Each image is published under an immutable `sha-<12-char-sha>` tag. That is
+# what makes rollback possible: `IMAGE_TAG=sha-abc123456789 ./deploy.sh`
+# redeploys an exact prior build. 12 characters rather than git's default 7:
+# these tags are the rollback handle, and a 7-char prefix collision would make
+# an operator's chosen tag silently ambiguous.
+#
+# `latest` is published by a SEPARATE `promote-latest` job, never by the build
+# matrix, for two reasons:
+#
+#   1. `workflow_dispatch` can be run from ANY branch. If the matrix pushed
+#      `latest` directly, a manual run from a feature branch would overwrite the
+#      tag that production pulls by default with unmerged code. `promote-latest`
+#      is gated on `refs/heads/master`.
+#   2. The matrix builds the two images as independent jobs. If `latest` were
+#      pushed per-job, a run where `server` succeeded and `internal-service`
+#      failed would leave `latest` pointing at a MISMATCHED pair, and a deploy
+#      using the default tag would pull two versions that never shipped
+#      together. `promote-latest` needs the whole matrix to succeed, so `latest`
+#      only ever advances as a complete set.
 #
 # ONE-TIME MANUAL SETUP
 #
@@ -58,8 +75,21 @@ jobs:
         include:
           - image: server
             dockerfile: ./docker/workspace-server/Dockerfile
+            # Smoke-test config. The server refuses to start without a master
+            # password, and we exercise the filesystem backend because that is
+            # what production runs.
+            smoke_port: 12349
+            smoke_env: >-
+              -e WORKSPACE_MASTER_PASSWORD=ci-smoke-test
+              -e WORKSPACE_BACKEND=filesystem
+              -e WORKSPACE_DATA_DIR=/data/server
+              -e WORKSPACE_BIND_ADDR=0.0.0.0:12349
           - image: internal-service
             dockerfile: ./docker/internal-service/Dockerfile
+            smoke_port: 12345
+            smoke_env: >-
+              -e INTERNAL_SERVICE_PORT=12345
+              -e INTERNAL_SERVICE_BIND_HOST=0.0.0.0
 
     steps:
       - uses: actions/checkout@v4
@@ -80,14 +110,16 @@ jobs:
           sudo apt-get clean
           df -h
 
+      - uses: docker/setup-buildx-action@v3
+
       - name: Compute image reference
         # Docker image references MUST be lowercase, but the org is spelled
         # `Avarok-Cybersecurity`. Passing ${{ github.repository_owner }} straight
         # into a tag produces `ghcr.io/Avarok-Cybersecurity/...`, which the daemon
-        # rejects with "repository name must be lowercase". Lowercase it here.
+        # rejects with "repository name must be lowercase".
         run: |
           echo "IMAGE=ghcr.io/${GITHUB_REPOSITORY_OWNER,,}/citadel-workspace-${{ matrix.image }}" >> "$GITHUB_ENV"
-          echo "SHA_TAG=sha-$(git rev-parse --short HEAD)" >> "$GITHUB_ENV"
+          echo "SHA_TAG=sha-${GITHUB_SHA:0:12}" >> "$GITHUB_ENV"
 
       - name: Log in to GHCR
         uses: docker/login-action@v3
@@ -96,7 +128,7 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build and push
+      - name: Build production image (load locally, do NOT push yet)
         uses: docker/build-push-action@v6
         with:
           context: .
@@ -106,19 +138,84 @@ jobs:
           # that docker-compose.yml (which pins no target) keeps working. Omitting
           # this line would publish the 5.6 GB toolchain image.
           target: production
-          push: true
+          # Load into the local daemon instead of pushing, so the smoke test below
+          # runs against the EXACT artifact that will be published. Pushing first
+          # and testing afterwards would leave a broken image in the registry for
+          # anyone who pulled in between.
+          load: true
+          push: false
           build-args: |
             INTERNAL_SERVICE_PORT=12345
-          tags: |
-            ${{ env.IMAGE }}:latest
-            ${{ env.IMAGE }}:${{ env.SHA_TAG }}
-          # No BuildKit `gha` cache on purpose: the builder stage is ~5.6 GB and
-          # the Actions cache is capped at 10 GB, so caching these layers evicts
-          # itself and produces flaky, slow builds. This job runs once per merge to
-          # master; a clean build is the reliable choice.
+          tags: ${{ env.IMAGE }}:${{ env.SHA_TAG }}
 
-      - name: Report published tags
+      - name: Smoke test the image before publishing
+        # The runtime stages deliberately omit the toolchain, the source tree and
+        # most OS packages. A missing shared library, a missing runtime asset
+        # (kernel.toml / documents/), or a broken CMD would otherwise only surface
+        # during an actual deploy. This starts the real image and probes it with
+        # `nc -z` - the same check the compose healthchecks use - so a broken image
+        # can never reach the registry.
         run: |
-          echo "Published:"
-          echo "  ${IMAGE}:latest"
-          echo "  ${IMAGE}:${SHA_TAG}   <- pin this to roll back"
+          set -euo pipefail
+          name="smoke-${{ matrix.image }}"
+          docker run -d --name "$name" ${{ matrix.smoke_env }} "$IMAGE:$SHA_TAG"
+
+          ok=0
+          for _ in $(seq 1 45); do
+            state=$(docker inspect -f '{{.State.Status}}' "$name" 2>/dev/null || echo gone)
+            if [ "$state" != "running" ]; then
+              echo "Container is '$state' - it exited instead of serving."
+              break
+            fi
+            if docker exec "$name" nc -z 127.0.0.1 ${{ matrix.smoke_port }} 2>/dev/null; then
+              ok=1
+              break
+            fi
+            sleep 2
+          done
+
+          if [ "$ok" -ne 1 ]; then
+            echo "::error::${{ matrix.image }} production image failed its smoke test - not publishing."
+            docker logs "$name" 2>&1 | tail -50 || true
+            exit 1
+          fi
+          echo "${{ matrix.image }} is listening on ${{ matrix.smoke_port }}; image is good."
+          docker rm -f "$name" >/dev/null 2>&1 || true
+
+      - name: Push immutable SHA tag
+        # Only the immutable tag is pushed here. `latest` is advanced by the
+        # `promote-latest` job below, and only when the whole matrix succeeded on
+        # master - see the header comment.
+        run: docker push "$IMAGE:$SHA_TAG"
+
+  promote-latest:
+    name: Promote latest
+    # `needs` requires EVERY matrix job to have succeeded. Combined with the
+    # master-only guard, this is what guarantees `latest` never points at a
+    # half-published release or at unmerged code from a manual branch dispatch.
+    needs: publish
+    if: github.ref == 'refs/heads/master'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: docker/setup-buildx-action@v3
+
+      - name: Log in to GHCR
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Point latest at this release
+        # `imagetools create` retags in the registry by manifest reference - it
+        # does not pull or rebuild the image, so the bits under `latest` are
+        # byte-identical to the SHA tag that just passed its smoke test.
+        run: |
+          set -euo pipefail
+          sha_tag="sha-${GITHUB_SHA:0:12}"
+          for img in server internal-service; do
+            image="ghcr.io/${GITHUB_REPOSITORY_OWNER,,}/citadel-workspace-${img}"
+            echo "Promoting ${image}:${sha_tag} -> ${image}:latest"
+            docker buildx imagetools create -t "${image}:latest" "${image}:${sha_tag}"
+          done
+          echo "Rollback handle for this release: IMAGE_TAG=${sha_tag}"
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -75,6 +75,11 @@ jobs:
         include:
           - image: server
             dockerfile: ./docker/workspace-server/Dockerfile
+            # The server Dockerfile declares no build args. Passing an ARG a Dockerfile does not
+            # declare is a silent no-op that Docker only warns about, so build args are scoped
+            # per-image here rather than passed globally to both.
+            build_args: ""
+
             # Smoke-test config. The server refuses to start without a master
             # password, and we exercise the filesystem backend because that is
             # what production runs.
@@ -93,6 +98,10 @@ jobs:
               -e WORKSPACE_BIND_ADDR=127.0.0.1:12349
           - image: internal-service
             dockerfile: ./docker/internal-service/Dockerfile
+            # Only this Dockerfile declares INTERNAL_SERVICE_PORT (it is baked into the image's
+            # ENV and expanded by its CMD at runtime).
+            build_args: |
+              INTERNAL_SERVICE_PORT=12345
             smoke_port: 12345
             # Loopback, for the same reason as the server above.
             smoke_env: >-
@@ -152,8 +161,7 @@ jobs:
           # anyone who pulled in between.
           load: true
           push: false
-          build-args: |
-            INTERNAL_SERVICE_PORT=12345
+          build-args: ${{ matrix.build_args }}
           tags: ${{ env.IMAGE }}:${{ env.SHA_TAG }}
           # Stamp the commit into the image itself. `latest` is a MUTABLE tag on two
           # INDEPENDENT registry repositories, and no registry offers an atomic
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -79,17 +79,25 @@ jobs:
             # password, and we exercise the filesystem backend because that is
             # what production runs.
             smoke_port: 12349
+            # Bind to LOOPBACK, matching the production default. Nothing is published out of
+            # the container anyway (the smoke run uses no `-p` and no host networking, so the
+            # listener is confined to the container's own network namespace), and the probe
+            # runs via `docker exec` inside that namespace - so loopback costs nothing. But
+            # this repository has a self-hosted runner, and a container that binds 0.0.0.0 is
+            # one `-p` or one `network_mode: host` away from putting a server with a known
+            # password on someone's LAN. Default to the narrow thing.
             smoke_env: >-
               -e WORKSPACE_MASTER_PASSWORD=ci-smoke-test
               -e WORKSPACE_BACKEND=filesystem
               -e WORKSPACE_DATA_DIR=/data/server
-              -e WORKSPACE_BIND_ADDR=0.0.0.0:12349
+              -e WORKSPACE_BIND_ADDR=127.0.0.1:12349
           - image: internal-service
             dockerfile: ./docker/internal-service/Dockerfile
             smoke_port: 12345
+            # Loopback, for the same reason as the server above.
             smoke_env: >-
               -e INTERNAL_SERVICE_PORT=12345
-              -e INTERNAL_SERVICE_BIND_HOST=0.0.0.0
+              -e INTERNAL_SERVICE_BIND_HOST=127.0.0.1
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -1,0 +1,124 @@
+# =============================================================================
+# Publish container images to GHCR.
+#
+# WHY THIS EXISTS
+#
+# Deploys previously built images ON the production host (`deploy.sh` ran
+# `docker compose build`). That is slow (a full Rust release build per deploy),
+# it requires the source tree and a toolchain on the production box, it offers
+# no rollback (the previous image is simply gone), and it makes every deploy
+# hostage to that host's Docker networking. Building here instead and having the
+# host `docker compose pull` fixes all four at once.
+#
+# Images are tagged with BOTH:
+#   * `sha-<short-sha>` - immutable. This is what makes rollback possible:
+#     `IMAGE_TAG=sha-abc1234 ./deploy.sh` redeploys an exact prior build.
+#   * `latest`          - moving pointer to the newest master build.
+#
+# ONE-TIME MANUAL SETUP
+#
+# GHCR packages are created PRIVATE. After the first successful run, open each
+# package's settings on GitHub and set its visibility to Public, otherwise every
+# developer must `docker login ghcr.io` before they can pull. The repositories
+# themselves are already public, so there is nothing secret in these images.
+#
+# WHY `ui` IS NOT PUBLISHED HERE (yet)
+#
+# The production UI image bakes `VITE_WS_URL` at build time, and its nginx CSP
+# is `connect-src 'self'` - so a UI served on :8080 cannot reach an agent on
+# :12345. Publishing it today would ship an artifact that cannot work for the
+# local-agent deployment. It gets added once the same-origin `/ws` reverse proxy
+# (and the matching frontend change to derive the URL from `window.location`)
+# lands.
+# =============================================================================
+name: Publish Images
+
+on:
+  push:
+    branches: ["master"]
+  workflow_dispatch:
+
+# Never cancel a publish mid-push: a half-pushed manifest is worse than a slow
+# one, and unlike `validate.yml` these runs have side effects outside the runner.
+concurrency:
+  group: publish-images-${{ github.ref }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  publish:
+    name: Publish ${{ matrix.image }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - image: server
+            dockerfile: ./docker/workspace-server/Dockerfile
+          - image: internal-service
+            dockerfile: ./docker/internal-service/Dockerfile
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Free Disk Space
+        # The `builder` stage is ~5.6 GB (full Rust toolchain + target/). A stock
+        # GitHub runner does not have room for it alongside the preinstalled
+        # toolchains, and the failure mode is an opaque "no space left on device"
+        # deep inside a cargo build. Same step validate.yml already uses.
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /opt/ghc
+          sudo rm -rf /usr/local/share/boost
+          sudo rm -rf /usr/local/.ghcup
+          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+          sudo apt-get clean
+          df -h
+
+      - name: Compute image reference
+        # Docker image references MUST be lowercase, but the org is spelled
+        # `Avarok-Cybersecurity`. Passing ${{ github.repository_owner }} straight
+        # into a tag produces `ghcr.io/Avarok-Cybersecurity/...`, which the daemon
+        # rejects with "repository name must be lowercase". Lowercase it here.
+        run: |
+          echo "IMAGE=ghcr.io/${GITHUB_REPOSITORY_OWNER,,}/citadel-workspace-${{ matrix.image }}" >> "$GITHUB_ENV"
+          echo "SHA_TAG=sha-$(git rev-parse --short HEAD)" >> "$GITHUB_ENV"
+
+      - name: Log in to GHCR
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: ${{ matrix.dockerfile }}
+          # `production` is the slim runtime stage (~145 MB). Without an explicit
+          # target, Docker builds the LAST stage - which is `dev`, deliberately, so
+          # that docker-compose.yml (which pins no target) keeps working. Omitting
+          # this line would publish the 5.6 GB toolchain image.
+          target: production
+          push: true
+          build-args: |
+            INTERNAL_SERVICE_PORT=12345
+          tags: |
+            ${{ env.IMAGE }}:latest
+            ${{ env.IMAGE }}:${{ env.SHA_TAG }}
+          # No BuildKit `gha` cache on purpose: the builder stage is ~5.6 GB and
+          # the Actions cache is capped at 10 GB, so caching these layers evicts
+          # itself and produces flaky, slow builds. This job runs once per merge to
+          # master; a clean build is the reliable choice.
+
+      - name: Report published tags
+        run: |
+          echo "Published:"
+          echo "  ${IMAGE}:latest"
+          echo "  ${IMAGE}:${SHA_TAG}   <- pin this to roll back"
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -147,6 +147,16 @@ jobs:
           build-args: |
             INTERNAL_SERVICE_PORT=12345
           tags: ${{ env.IMAGE }}:${{ env.SHA_TAG }}
+          # Stamp the commit into the image itself. `latest` is a MUTABLE tag on two
+          # INDEPENDENT registry repositories, and no registry offers an atomic
+          # multi-repository tag update -- so a promotion that succeeds for `server` and
+          # then fails for `internal-service` would leave `latest` pointing at a mismatched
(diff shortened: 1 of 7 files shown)


Tjemmmic:

> **Resolution (f64f203):** 4 addressed · 2 not an issue · 1 acknowledged. 

### Summary

The PR adds a GHCR publishing workflow, switches production deploys to pull prebuilt slim images, and introduces multi-stage Dockerfiles for server and internal-service. The approach is sound, but there are critical release-safety gaps: missing error handling in deploy.sh for pull/build failures that can cause silent partial deploys, non-atomic `latest` tag updates across matrix jobs, and no CI smoke test for the new production images.

### Issues Found
**7 total** — 2 P1 (blocking) · 4 P2 (should fix) · 1 P3 (nice to have)

#### ~~❗ P1 — docker compose pull failure does not abort deploy.sh~~ ❌ **Not an issue**

> **Note:** `deploy.sh:21` is `set -euo pipefail`, so a failing `docker compose pull` aborts the script - it cannot fall through to step 3. Verified empirically. The stated failure mode (silent partial deploy) cannot occur. An explicit, actionable message was added anyway, because a bare `set -e` abort prints nothing and the most likely cause here is a 403 from the GHCR packages still being Private. (f64f203)

- **File:** `deploy.sh:L191`
- **Problem:** The `docker compose pull` command has no error checking. If the pull fails (network issue, GHCR outage, image not found, authentication expired), the script proceeds to step 3 and runs `docker compose up -d` which will either use a stale local image or fail at runtime. Since the whole point of this PR is reliability, a silent pull failure that proceeds to restart services with old or missing images is a regression.
- **Fix:** Check the exit code of the pull command and abort if it fails:
```bash
if ! docker compose -f "$COMPOSE_FILE" "${PROFILE_ARGS[@]}" pull server internal-service; then
    echo "ERROR: Failed to pull images from GHCR. Aborting deploy." >&2
    exit 1
fi
```

---

#### ~~❗ P1 — docker compose build ui failure does not abort deploy.sh~~ ❌ **Not an issue**

> **Note:** Same reason. `set -e` exempts only the *condition* of an `if`, not commands in its body - verified: `bash -c 'set -e; if true; then false; fi; echo X'` never prints X. The build cannot silently continue. Explicit messaging added regardless. (f64f203)

- **File:** `deploy.sh:L196`
- **Problem:** The conditional `docker compose build ui` has no error handling. If the UI build fails (source tree missing on production host, build error), the script continues to step 3 and restarts server and internal-service with NEW images while `ui` still runs the OLD image or fails to build entirely. This leaves the deployment in a split-brain state where the UI is incompatible with the freshly-pulled backend images.
- **Fix:** Add explicit error handling around the ui build:
```bash
if ! docker compose -f "$COMPOSE_FILE" "${PROFILE_ARGS[@]}" build ui; then
    echo "ERROR: Failed to build ui image. Aborting deploy." >&2
    exit 1
fi
```

---

#### ~~⚠️ P2 — Manual workflow dispatch can publish branch builds as `latest`~~ ✅ **Addressed**

> **Note:** Real release-safety bug. The build matrix now pushes ONLY the immutable SHA tag; `latest` is advanced by a separate `promote-latest` job gated on `github.ref == 'refs/heads/master'`, so a manual dispatch from a feature branch can no longer overwrite the tag production pulls by default. (f64f203)

- **File:** `.github/workflows/publish-images.yml:L35-L36`
- **Problem:** `workflow_dispatch` is available on any branch, but the job always pushes both `latest` and `sha-<short>` tags. Running this workflow manually from a draft or feature branch would overwrite the production `latest` image with unmerged code, contradicting the documented contract that `latest` is the newest master build and causing production deploys with the default `IMAGE_TAG` to pull unmerged code.
- **Fix:** Only push `latest` when `github.ref == 'refs/heads/master'`. For manual branch dispatches, push only the immutable SHA tag. For example, conditionally include `latest` in the tags list based on the ref.

---

#### ~~⚠️ P2 — Default `latest` rollout can pull non-atomic server/internal-service versions~~ ✅ **Addressed**

> **Note:** Also real. `promote-latest` declares `needs: publish`, so `latest` only advances after EVERY image in the release has built and passed its smoke test. A run where one matrix job fails now leaves `latest` untouched rather than skewed across two versions. (f64f203)

- **File:** `docker-compose.production.yml:L49-L112`
- **Problem:** Both production services default to `${IMAGE_TAG:-latest}`, while the workflow publishes `latest` independently in matrix jobs for `server` and `internal-service` with `fail-fast: false`. A deploy that runs while only one matrix job has pushed, or after one succeeds and the other fails, can pull mismatched versions of the two services even though the deploy appears to target a single release.
- **Fix:** Make production deploys use an immutable tag by default or require `IMAGE_TAG` for production. For example, have CI emit a single release tag only after both images are built successfully, and make `deploy.sh` fail unless `IMAGE_TAG` is set to that immutable tag. Alternatively, remove `latest` from the production compose defaults and document the required `sha-<short>` tag.

---

#### ~~⚠️ P2 — No automated smoke test covers the new production image/deploy path~~ ✅ **Addressed**

> **Note:** The matrix now builds with `load: true, push: false`, starts the real image, and probes it with `nc -z` - the same check the compose healthchecks use - before pushing. A missing shared library, missing runtime asset or broken CMD now fails the job, so a broken image can never reach the registry. Testing before the push (rather than after) also means no one can pull a bad image in the interim. (f64f203)

- **File:** `.github/workflows/publish-images.yml:L103-L117`
- **Problem:** This PR replaces production host builds with pulled slim images and adds new runtime stages that omit the toolchain, source tree, build cache volume, and most OS packages. There is no test or workflow step that starts the `production` target images and verifies the existing compose healthchecks pass, so missing runtime assets/libs or a broken CMD would only be discovered during deployment. This is significant new deployment logic with no added validation.
- **Fix:** Add a CI smoke step before pushing, or a separate validation job, that builds the `production` targets and runs the services long enough for their healthchecks or ports to pass. At minimum, run `docker compose -f docker-compose.production.yml build server internal-service` followed by `docker compose -f docker-compose.production.yml up -d server internal-service` with required env vars and assert both services become healthy/reachable.

---

#### ~~⚠️ P2 — Image tags use only 7-char short SHA, risking collision and ambiguous rollback~~ ✅ **Addressed**

> **Note:** Now 12 characters (`${GITHUB_SHA:0:12}`) in both the publish and promote jobs, and the compose/deploy docs were updated to match. (f64f203)

- **File:** `.github/workflows/publish-images.yml:L62`
- **Problem:** `git rev-parse --short HEAD` defaults to 7 characters. For a rollback tool that relies on unambiguous tag identification, 7 chars has a non-trivial collision probability as the repo grows. If two SHAs share the same 7-char prefix, `IMAGE_TAG=sha-abc1234 ./deploy.sh` becomes ambiguous and could pull the wrong image.
- **Fix:** Use a longer short SHA for the tag to reduce collision risk:
```bash
echo "SHA_TAG=sha-$(git rev-parse --short=12 HEAD)" >> "$GITHUB_ENV"
```

---

#### ~~ℹ️ P3 — target_cache volume removed from production compose but orphans on prior deploys~~ ℹ️ **Acknowledged**

> **Note:** The concern is valid, but the suggested fix is unsafe and was deliberately not taken. Volume names are scoped by compose PROJECT name, so the orphan's name depends on the directory the stack was deployed from - and on the current deployment host a volume literally named `citadel-workspace_target_cache` belongs to the CI runner's own compose stack. A blind `docker volume rm` in deploy.sh could destroy an unrelated project's build cache. Documented the manual cleanup (check what owns it, then remove) instead of automating a destructive command. (f64f203)

- **File:** `docker-compose.production.yml:L286`
- **Problem:** The `target_cache` volume is removed from the compose file, but on hosts that previously ran the old compose file, the named volume still exists in Docker's volume store as an orphan. It is harmless but accumulates disk waste. The comment notes it is gone but does not address cleanup of pre-existing volumes.
- **Fix:** Add a note in the deploy script or migration documentation to remove the orphaned volume:
```bash
# Optional cleanup of orphaned volume from previous deploys
docker volume rm citadel-workspace_target_cache 2>/dev/null || true
```

---

> **Status:** Both P1 blockers were incorrect (`set -euo pipefail` already aborts the deploy; verified). All 4 P2s are fixed. The only remaining item is an acknowledged nit whose suggested auto-`docker volume rm` was rejected as unsafe on a shared host.

Two P1 issues around missing error handling for `docker compose pull` and `docker compose build ui` could cause silent partial deploys or restarts with stale/incompatible images, directly undermining the reliability goals of this PR. Additionally, there are 5 P2 issues including release-safety risks around the `latest` tag and missing test coverage for the new production image and deploy path, which taken together exceed the threshold for CHANGES_REQUESTED.

### Quick Reference
- **P1**: docker compose pull failure does not abort deploy.sh
- **P1**: docker compose build ui failure does not abort deploy.sh
- **P2**: Manual workflow dispatch can publish branch builds as `latest`
- **P2**: Default `latest` rollout can pull non-atomic server/internal-service versions
- **P2**: No automated smoke test covers the new production image/deploy path
- **P2**: Image tags use only 7-char short SHA, risking collision and ambiguous rollback
- **P3**: target_cache volume removed from production compose but orphans on prior deploys

---

Tjemmmic:

> **Resolution (a764ba9):** 1 addressed. 

### Summary

The PR moves production deploys to prebuilt, slim GHCR images with a well-architected publish workflow, but contains a release-consistency flaw. The mutable `latest` tag is advanced via sequential registry mutations rather than atomically, risking partial promotion and mismatched production images during a deploy.

### Issues Found
**1 total** — 1 P1 (blocking) · 0 P2 (should fix) · 0 P3 (nice to have)

#### ~~❗ P1 — `latest` promotion is not atomic across the two production images~~ ✅ **Addressed**

> **Note:** Correct, and unfixable at the registry: `latest` is a mutable tag on two INDEPENDENT repositories and no registry offers an atomic multi-repository tag update, so a partial promotion can always leave a mismatched pair. Fixed at the deploy end instead, per the third suggested remedy. Each image is now stamped at build time with `org.opencontainers.image.revision`, and `deploy.sh` reads that label from both pulled images and ABORTS before restarting anything if they disagree, printing both revisions and telling the operator to pin an immutable `IMAGE_TAG=sha-...`. Verified against real images: a matching pair proceeds, a mismatched pair aborts, and an unlabelled image (older or locally built) warns without blocking. This catches a partial promotion, a hand-edited tag, or an interrupted deploy alike - the promote job remains the happy path, this is the safety net. (a764ba9)

- **File:** `.github/workflows/publish-images.yml:L214-L217`
- **Problem:** The `promote-latest` job advances `server:latest` and `internal-service:latest` in a sequential loop. If the first `imagetools create` succeeds but the second fails (due to a transient registry/auth/network error), `latest` will point to a mismatched pair. Because `docker-compose.production.yml` defaults both services to `${IMAGE_TAG:-latest}`, a standard `./deploy.sh` will pull and restart an untested combination of versions, violating the PR's stated contract that `latest` only advances as a complete set.
- **Fix:** Do not make production deploys rely on independently mutable `latest` tags for multi-image releases. Require an immutable `IMAGE_TAG=sha-...` for production deploys, publish a single release manifest/metadata file recording the exact image digests for `deploy.sh` to pull, or have `deploy.sh` inspect the pulled images' revision labels and abort if they belong to different commits.

---

> **Status:** The sole P1 is fixed: production can no longer be restarted on a mismatched image pair, because the deploy verifies both images' commit revision labels and aborts if they differ.

Two of the three independent reviews flagged a release-safety issue where a failed or partial promotion can leave the default deploy target (`latest`) pointing at a mixed pair of images. Because this can ship mismatched backend components to production during a normal deploy or transient CI failure, it constitutes a blocking issue that must be fixed before merging.

### Quick Reference
- **P1**: `latest` promotion is not atomic across the two production images

---

Tjemmmic:

> **Resolution (7188056):** 1 addressed · 4 not an issue. 

### Summary

This PR introduces a CI workflow to build and publish slim Docker images to GHCR, alongside updates to `deploy.sh` to pull these prebuilt images instead of compiling on the host. While the operational logic and extensive comments are excellent, there is a critical build failure in both production Dockerfiles due to an incorrect package name for OpenSSL on `debian:trixie-slim`.

### Issues Found
**5 total** — 2 P1 (blocking) · 3 P2 (should fix) · 0 P3 (nice to have)

#### ~~❗ P1 — Workspace server production image installs the wrong OpenSSL package for trixie~~ ❌ **Not an issue**

> **Note:** The build does not fail. `libssl3` is a VIRTUAL package on trixie, provided by `libssl3t64` (`apt-cache showpkg libssl3` -> `Reverse Provides: libssl3t64 3.5.6-1~deb13u2`), so `apt-get install libssl3` resolves and exits 0 - verified by running the Dockerfile's exact apt line in a clean `debian:trixie-slim` container, and independently by the fact that both production images were already built and run successfully with it (`ldd` resolves libssl.so.3 and libcrypto.so.3). That said, the underlying instinct is right: depending on a virtual alias is less explicit than naming the real package, and an alias can be dropped in a future point release. Switched both Dockerfiles to `libssl3t64` and re-verified end-to-end - both images build, are unchanged in size (142 MB / 145 MB), still carry libssl.so.3, and the server boots, seeds and passes its `nc -z` probe with zero errors. (7188056)

- **File:** `docker/workspace-server/Dockerfile:L70-L80`
- **Problem:** The production runtime stage is based on `debian:trixie-slim` but installs `libssl3`. In Debian trixie, the shared-library package is `libssl3t64`. `apt-get install` will fail, blocking the GHCR publish job.
- **Fix:** Change `libssl3` to `libssl3t64` in the production-stage package list.

---

#### ~~❗ P1 — Internal service production image installs the wrong OpenSSL package for trixie~~ ❌ **Not an issue**

> **Note:** Same as above - the build succeeds, `libssl3` is a virtual package provided by `libssl3t64`. Switched to the real package name anyway and re-verified the image builds and runs. (7188056)

- **File:** `docker/internal-service/Dockerfile:L70-L80`
- **Problem:** This production stage has the same `debian:trixie-slim` plus `libssl3` mismatch. The internal-service image build will fail during `apt-get install`, blocking the publish matrix.
- **Fix:** Change `libssl3` to `libssl3t64` in the production-stage package list.

---

#### ~~⚠️ P2 — deploy.sh introduces an unchecked jq dependency on production hosts~~ ❌ **Not an issue**

> **Note:** `deploy.sh` already preflights `jq` at lines 92-97, long before step 2 pulls anything - it exits with an actionable install hint if `jq` is missing. The dependency is not new either; the readiness probe has always used it. (no change needed)

- **File:** `deploy.sh:L229-L230`
- **Problem:** The new release-consistency gate parses `docker compose config --format json` with `jq`, but the deploy script does not preflight that `jq` exists. On a production host without `jq`, the script will abort after pulling images with a shell error instead of intended deploy diagnostics.
- **Fix:** Add an early preflight check before pulling images: `command -v jq >/dev/null || { echo "ERROR: jq is required by deploy.sh" >&2; exit 1; }`.

---

#### ~~⚠️ P2 — Production image workflow only runs after merge~~ ❌ **Not an issue**

> **Note:** The production targets ARE validated at PR time. `validate.yml`'s `production-build` job builds `docker-compose.production.yml` - which this PR pins to `target: production` - and smoke-tests both services on every pull request. So a broken slim Dockerfile fails the PR, not the post-merge publish. Adding a duplicate `pull_request` build to this workflow would double a ~15-minute Rust build for coverage that already exists. (What genuinely cannot be exercised pre-merge is the login/push/promote path itself, which is inherent to publishing.)

- **File:** `.github/workflows/publish-images.yml:L54-L58`
- **Problem:** The new workflow is triggered only on `push` to `master` and `workflow_dispatch`. Significant new packaging logic and slim Dockerfiles are untested in PRs, meaning a broken production target can land on `master` and only fail in the post-merge publish/deploy path.
- **Fix:** Add a `pull_request` validation path that builds the same `production` targets and runs the same smoke tests without logging in or pushing. Keep the push/promote steps conditional on `github.event_name != 'pull_request'`.

---

#### ~~⚠️ P2 — CI smoke test cleanup lacks a trap~~ ✅ **Addressed**

> **Note:** Valid, and it matters more here than the review could know: this repository has a SELF-HOSTED runner, where a container leaked by a failed run keeps its ports bound and poisons the next run - a five-month-old leaked compose stack squatting on :12345/:12349 is exactly how the deployment host ended up blocked. Added `trap 'docker rm -f "$name" || true' EXIT` so cleanup happens on every exit path. (7188056)

- **File:** `.github/workflows/publish-images.yml:L107-L124`
- **Problem:** The smoke test step exits with `exit 1` on failure but only runs `docker rm -f "$name"` on success (line 124). If the test fails, the container is left running, which can cause issues on self-hosted runners.
- **Fix:** Use a `trap` to ensure the container is always cleaned up, regardless of test outcome: `trap 'docker rm -f "$name" >/dev/null 2>&1 || true' EXIT`.

---

> **Status:** Neither P1 is a build failure: `libssl3` is a virtual package provided by `libssl3t64`, proven by running the apt line directly and by two successful image builds. The real package name was adopted anyway and re-verified. `jq` is already preflighted (deploy.sh:92), and the production targets are already PR-validated by validate.yml's production-build. The trap is added.

The P1 package name mismatch (`libssl3` vs `libssl3t64`) is a release-blocking build failure for both production images. Additionally, the unchecked dependency on `jq` poses an unnecessary risk to production deploys, and the new packaging logic desperately needs PR-time validation. These issues must be resolved before merging to ensure production stability.

### Quick Reference
- **P1**: Workspace server production image installs the wrong OpenSSL package for trixie
- **P1**: Internal service production image installs the wrong OpenSSL package for trixie
- **P2**: deploy.sh introduces an unchecked jq dependency on production hosts
- **P2**: Production image workflow only runs after merge
- **P2**: CI smoke test cleanup lacks a trap

---

Tjemmmic:

> **Resolution (da5b201):** 2 addressed · 1 not an issue. 

### Summary

The PR moves production backend deployment from host-side builds to CI-published GHCR images and adds a release-consistency gate before restart. The direction is sound and well-documented, but the new deploy control flow introduces an unguarded `jq` host dependency and has zero test coverage for significant production deployment logic.

### Issues Found
**3 total** — 0 P1 (blocking) · 2 P2 (should fix) · 1 P3 (nice to have)

#### ~~⚠️ P2 — Deploy now requires jq without a preflight check or documented install requirement~~ ❌ **Not an issue**

> **Note:** `jq` is already preflighted at `deploy.sh:92-97`, which runs long before step 2 pulls anything: it exits with an actionable install hint (`apt-get install jq` / `brew install jq`) if `jq` is absent. So the described failure - pulling images and then dying mid-deploy on `jq: command not found` - cannot occur; the script never reaches the pull. The dependency is not new either: the readiness probe has used `jq` since before this PR. (no change needed)

- **File:** `deploy.sh:L230-L231`
- **Problem:** The new consistency gate unconditionally pipes `docker compose config --format json` into `jq`. Production hosts that previously only needed Docker Compose can now fail mid-deploy after successfully pulling images but before restart, with `jq: command not found`. This is a new host dependency introduced in the critical deployment path without detection or a clear actionable error.
- **Fix:** Add an explicit preflight check before the pull/restart sequence. For example, check `command -v jq >/dev/null` near the existing dependency checks and fail with an installation message before any image pull side effects occur.

---

#### ~~⚠️ P2 — New deploy pull and revision gate has no regression coverage~~ ✅ **Addressed**

> **Note:** Agreed, and this was the right thing to insist on - the revision gate is the mechanism that stops a partially promoted `latest` from restarting production on a mismatched pair, and it was the one part of the deploy path with no test. Extracted it into `scripts/verify-image-revisions.sh`, which takes images as arguments and touches no global state, and added a `deploy-gate-tests` CI job that exercises it against REAL images (a few `FROM busybox` layers with real labels, so it tests `docker image inspect` behaviour rather than a mock of it - including the literal `<no value>` string docker prints for an absent label, which a stub would paper over). The job pins all four paths: matching revisions deploy, mismatched revisions abort, unlabelled images warn without blocking, and both scripts parse. (da5b201)

- **File:** `deploy.sh:L196-L262`
- **Problem:** This block is significant new production deployment logic: it pulls remote images, derives the effective image names from compose JSON, reads OCI revision labels, and decides whether production is allowed to restart. The PR adds no test or CI check that exercises the success path, mismatched-label abort, unlabeled-image fallback, or failed-pull behavior, so a shell/compose/jq regression here would only be discovered during an actual deploy.
- **Fix:** Add a focused deploy-script test or CI job that stubs `docker`/`docker compose`/`jq` and verifies at least: pull failure exits before any `up`, matching labels allow the restart phase, mismatched labels exit 1 before restart, missing labels take the documented fallback path, and UI build failure exits before backend restart.

---

#### ~~ℹ️ P3 — Deploy prints a false commit confirmation after skipping the consistency check~~ ✅ **Addressed**

> **Note:** Correct - on the unlabelled path it warned that the check was skipped and then printed `Both images are from commit ` with an empty revision, which flatly contradicts the warning and would mislead an operator mid-rollback. The success line now only prints when every image was actually verified; the skip path prints `Revision check skipped: at least one image carries no revision label.` and nothing else. Pinned by the `deploy-gate-tests` job. (da5b201)

- **File:** `deploy.sh:L233-L254`
- **Problem:** When either image has no `org.opencontainers.image.revision` label, lines 233-237 explicitly skip the consistency check, but line 254 still prints `Both images are from commit ${srv_rev}`. For unlabeled images this emits an empty or `<no value>` commit and contradicts the warning, which can mislead an operator during rollback or first deployment from older/local images.
- **Fix:** Move the success message into the matching-label branch only. For example, after the mismatch check, print the commit only in an `else` branch where both labels are present and equal; do not print it after the warning path.

---

> **Status:** The `jq` concern is incorrect (it is preflighted at deploy.sh:92, before any pull). The revision gate is now extracted into a standalone script with a CI job covering matching / mismatched / unlabelled paths against real images, and the false 'verified' message on the skip path is gone.

The PR introduces substantial new production deployment logic with an unguarded `jq` dependency and no regression tests for the pull/revision gate. Per the decision rules, two P2s do not mandate changes, but the untested deployment logic combined with the potential for mid-deploy failure on hosts lacking `jq` presents a concrete reliability regression. Addressing the preflight check and adding tests for the deploy control flow will ensure the safety mechanisms behave as intended.

### Quick Reference
- **P2**: Deploy now requires jq without a preflight check or documented install requirement
- **P2**: New deploy pull and revision gate has no regression coverage
- **P3**: Deploy prints a false commit confirmation after skipping the consistency check

---

Tjemmmic:

> **Resolution (ef97481):** 3 addressed · 2 not an issue. 

### Summary

The PR adds GHCR image publishing and updates deploy.sh to pull prebuilt images instead of building on the host. Architecturally sound and well-documented, but the new release-consistency safety gate in verify-image-revisions.sh is easily bypassed: it treats missing images (from Docker inspect failures) the same as unlabelled images, and the CI regression test actively locks in this unsafe behavior instead of catching it.

### Issues Found
**5 total** — 0 P1 (blocking) · 5 P2 (should fix) · 0 P3 (nice to have)

#### ~~⚠️ P2 — Revision gate silently passes when Docker cannot inspect an image~~ ✅ **Addressed**

> **Note:** Right, and this was the worst kind of bug: the safety gate had a hole exactly where it mattered. `image_revision()` swallowed an inspect failure via `|| true` and returned an empty revision, so a MISSING image - a half-completed `docker compose pull`, a typo'd tag, a dead daemon - took the warn-and-continue path reserved for intentionally unlabelled local images and bypassed the gate entirely. It now distinguishes the two: a failed `docker image inspect` returns non-zero and aborts the deploy with a specific message, while a successful inspect with an absent label still warns and continues. An image we cannot see is an image we cannot verify. (ef97481)

- **File:** `scripts/verify-image-revisions.sh:36-43`
- **Problem:** image_revision() converts any docker image inspect failure into an empty revision via `|| true`. A missing image, bad reference, or Docker daemon error is treated identically to an intentionally unlabeled local image, completely bypassing the release-consistency gate. If a `docker compose pull` partially fails, deploy.sh proceeds and relies on `docker compose up` to fail later, defeating the purpose of this explicit safety check.
- **Fix:** Differentiate between an inspect failure and an absent label. Return a non-zero exit code from image_revision() if `docker image inspect` fails, allowing the warn-and-continue path only for successfully inspected images with an absent label.

---

#### ~~⚠️ P2 — Missing-image regression test passes even when the gate allows the missing image~~ ✅ **Addressed**

> **Note:** Fair, and the more damaging of the two: the step was named 'Missing image is refused' but passed whichever way the script behaved, so it PINNED the bypass instead of catching it - a test that cannot fail is worse than no test, because it advertises coverage that does not exist. The assertion is now inverted: if the gate allows an image it could not inspect, the job fails with an explicit error. Verified locally against all four paths - matching allowed, mismatched refused, unlabelled-but-inspectable warned and allowed, missing refused. (ef97481)

- **File:** `.github/workflows/validate.yml:429-441`
- **Problem:** The step is named 'Missing image is refused', but if verify-image-revisions.sh exits 0 for a missing image, the script only prints a warning and still exits 0. This pins the unsafe bypass behavior in CI instead of catching it.
- **Fix:** Invert the assertion so that success is a test failure: `if ./scripts/verify-image-revisions.sh rev-a definitely-not-pulled:latest 2>/dev/null; then echo "::error::missing image was allowed"; exit 1; fi`.

---

#### ~~⚠️ P2 — Deploy script adds an undeclared jq dependency on production hosts~~ ❌ **Not an issue**

> **Note:** `jq` is preflighted at `deploy.sh:92-97` - a top-level, unconditional check that runs long before step 2 pulls anything and exits with an install hint (`apt-get install jq` / `brew install jq`). The described failure - pulling images and then dying on `command not found` - cannot occur, because the script exits before the pull. Nor is the dependency new: the readiness probe (`wait_for_port`) has parsed `docker compose ps` with `jq` since before this PR. (no change needed)

- **File:** `deploy.sh:226-227`
- **Problem:** The new release gate uses `jq` to extract image names from the compose config, but deploy.sh previously did not require jq and has no preflight check or fallback. On a minimal production host without jq, the deploy will fail with a command-not-found error after images have already been pulled.
- **Fix:** Add a preflight check early in deploy.sh: `command -v jq >/dev/null || { echo "ERROR: jq is required" >&2; exit 1; }` or avoid jq by using native Docker Compose formatting options.

---

#### ~~⚠️ P2 — Smoke test exposes the server to the internet with a hardcoded password~~ ✅ **Addressed**

> **Note:** The exposure as described does not occur: the smoke `docker run` uses no `-p` and no `--network=host`, so a 0.0.0.0 bind is confined to the container's own network namespace and reaches neither the host nor the LAN, and the probe runs via `docker exec` inside that namespace. But the underlying instinct is right and the change is free, so both smoke tests now bind loopback, matching the production default. A container that binds 0.0.0.0 is one `-p` away from putting a server with a known password on a self-hosted runner's network; defaulting to the narrow thing costs nothing. (ef97481)

- **File:** `.github/workflows/publish-images.yml:45`
- **Problem:** The server smoke test runs with `WORKSPACE_BIND_ADDR=0.0.0.0:12349`. While GitHub-hosted runners are ephemeral, this repository also uses a self-hosted runner. Binding to 0.0.0.0 on a self-hosted runner unnecessarily exposes the test server and its hardcoded `ci-smoke-test` password to the local network during the ~90-second test window.
- **Fix:** Bind the smoke test to loopback to match the production default and reduce the attack surface: change `WORKSPACE_BIND_ADDR=0.0.0.0:12349` to `WORKSPACE_BIND_ADDR=127.0.0.1:12349` in the smoke_env for the server image.

---

#### ~~⚠️ P2 — No PR-time validation for the new production Dockerfile stages~~ ❌ **Not an issue**

> **Note:** The production stages ARE validated on every pull request. `docker-compose.production.yml` pins `target: production` for both images (lines 65 and 125), and `validate.yml`'s existing `production-build` job builds from that compose file and then smoke-tests both services - so a missing runtime shared library fails the PR, not the post-merge publish. Adding a second PR-time build of the same targets to the publish workflow would duplicate a ~15-minute Rust build for coverage that already exists. (no change needed)

- **File:** `docker/workspace-server/Dockerfile:60-130`
- **Problem:** The PR introduces new multi-stage Dockerfiles with a `production` stage relying on manually specified runtime dependencies derived from `ldd` output. The publish workflow only builds this target on pushes to master, so a dependency change adding a new shared library could merge via a PR and only break the image when it reaches master.
- **Fix:** Add a job to validate.yml that builds the `production` target for both images on PRs, ensuring that any missing runtime dependencies are caught before merging.

---

> **Status:** The gate's real hole is closed - it no longer treats an un-inspectable image as merely unlabelled, and the CI step now asserts that refusal instead of pinning the bypass. Smoke tests bind loopback. `jq` is already preflighted (deploy.sh:92) and the production stages are already PR-validated by validate.yml's production-build (compose pins `target: production`).

The architectural shift to prebuilt GHCR images is a strong improvement, but the core safety gate meant to protect production deploys is bypassed by missing images and its CI test locks in the unsafe behavior. Combined with the undeclared jq dependency and lack of PR-time validation for the new production Dockerfile stages, the PR introduces significant release-path risks that must be addressed before merging.

### Quick Reference
- **P2**: Revision gate silently passes when Docker cannot inspect an image
- **P2**: Missing-image regression test passes even when the gate allows the missing image
- **P2**: Deploy script adds an undeclared jq dependency on production hosts
- **P2**: Smoke test exposes the server to the internet with a hardcoded password
- **P2**: No PR-time validation for the new production Dockerfile stages

---

Tjemmmic:

> **Resolution:** 2 addressed · 1 not an issue. 

### Summary

This PR introduces GHCR prebuilt images and moves production deployment to a pull-based flow, alongside a deploy-time revision consistency gate. While the architectural approach is sound, the release path has a critical safety regression: it rebuilds the UI after backend services have already been restarted, which defeats the script's own preflight build guard and leaves the system vulnerable to partial updates.

### Issues Found
**3 total** — 1 P1 (blocking) · 2 P2 (should fix) · 0 P3 (nice to have)

#### ~~❗ P1 — UI is rebuilt again after backend restart, allowing a late UI build failure to leave production partially updated~~ ✅ **Addressed**

> **Note:** Correct, and it defeated the whole point of the preflight. Step 2 builds the ui precisely so a build failure aborts with the old stack untouched - and then step 3 rebuilt it with `up -d --no-deps --build ui` AFTER the server and internal-service had already been swapped, so a failure at that point would leave production on a new backend with the old UI: the mixed-version state the preflight exists to prevent. Dropped `--build` from the ui restart; all three restarts are now build-free and everything is built before anything is touched. (b94acbe)

- **File:** `deploy.sh:L244-L348`
- **Problem:** The new preflight block explicitly attempts to build the `ui` before restarting anything so that a build failure aborts with the old stack untouched. However, the actual restart sequence still uses `docker compose ... up -d --no-deps --build ui` after the `server` and `internal-service` images have already been restarted. If that second build invocation fails due to cache invalidation, disk pressure, or transient issues, production is left with new backend images and the old UI, exactly the mixed-version state the preflight block claims to prevent.
- **Fix:** Since the UI image is successfully built once in the preflight block, restart it without rebuilding. Remove the `--build` flag from the UI restart command:
```bash
# UI last (lightweight, fast restart)
echo "  Restarting ui..."
docker compose -f "$COMPOSE_FILE" "${PROFILE_ARGS[@]}" up -d --no-deps ui
```

---

#### ~~⚠️ P2 — Deploy script requires jq but does not validate the dependency~~ ❌ **Not an issue**

> **Note:** `jq` is preflighted at `deploy.sh:92-97` - a top-level, unconditional check that runs before step 2 pulls anything and exits with an install hint. The described failure (pull succeeds, then the script dies on `jq: command not found` before the restart) cannot occur, because the script exits long before the pull. The dependency also predates this PR: `wait_for_port` has parsed `docker compose ps` with `jq` all along. (no change needed)

- **File:** `deploy.sh:L230-L232`
- **Problem:** The release-consistency gate pipes `docker compose config --format json` directly into `jq` to extract the image reference, but the script does not check if `jq` is installed. On a production host without `jq`, the image pull succeeds but the script will abort mid-deployment with a shell-level `jq: command not found`, right before the restart sequence.
- **Fix:** Add a check for `jq` near the top of the script or alongside the other deploy preflight checks, failing fast if it is missing:
```bash
if ! command -v jq >/dev/null 2>&1; then
    echo "ERROR: jq is required to verify pulled image revisions." >&2
    exit 1
fi
```

---

#### ~~⚠️ P2 — Hardcoded INTERNAL_SERVICE_PORT in workflow build-args~~ ✅ **Addressed**

> **Note:** Correct - the workspace-server Dockerfile declares no build args, so passing `INTERNAL_SERVICE_PORT` to it was a silent no-op that Docker merely warns about. Build args are now scoped per-image in the matrix: only the internal-service entry carries it (it is the only Dockerfile that declares the ARG, bakes it into ENV, and expands it in its CMD). (b94acbe)

- **File:** `.github/workflows/publish-images.yml:L100-L101`
- **Problem:** The build step passes `INTERNAL_SERVICE_PORT=12345` as a build-arg for both the server and internal-service images via the matrix workflow. The workspace-server Dockerfile does not accept this build arg, making it a no-op for the server build that introduces configuration drift if the port ever changes.
- **Fix:** Move the `INTERNAL_SERVICE_PORT` build-arg out of the global workflow step and into the matrix configuration exclusively for the internal-service entry, or omit it entirely and let the Dockerfile/compose defaults handle it.

---

> **Status:** The P1 is fixed - the ui is no longer rebuilt after the backends are restarted, so a late build failure can no longer leave production half-updated. Build args are scoped per-image. `jq` is already preflighted at deploy.sh:92.

The PR's core contribution is a safer pull-and-restart deployment flow, but the deployment script contains a critical regression where a UI rebuild failure after backend restarts leaves the system in a mixed-version state. Combined with an undocumented `jq` dependency and hardcoded build arguments, there are enough operational risks to block this PR until the deployment safety logic is corrected.

### Quick Reference
- **P1**: UI is rebuilt again after backend restart, allowing a late UI build failure to leave production partially updated
- **P2**: Deploy script requires jq but does not validate the dependency
- **P2**: Hardcoded INTERNAL_SERVICE_PORT in workflow build-args

---

Tjemmmic:

> **Resolution:** 2 not an issue. 

### Summary

This PR introduces a robust GHCR publish workflow with smoke tests, slim production Docker images, and a deploy-time revision-consistency gate, moving production away from host builds. While the architecture and safety gates are well-designed and tested, the deploy path now depends on `jq` without explicitly verifying or documenting it. Additionally, the server smoke test may fail at startup due to missing required environment variables.

### Issues Found

#### ~~⚠️ P2 — deploy.sh introduces undeclared jq dependency without a preflight check~~ ❌ **Not an issue**

> **Note:** `jq` is preflighted at `deploy.sh:92-97` - a top-level, unconditional check that runs BEFORE step 2 pulls anything and exits with an install hint. The described sequence (images pulled, then the script aborts on `jq` before the restart) cannot occur, because the script exits long before the pull. The dependency also predates this PR: `wait_for_port` has parsed `docker compose ps` with `jq` all along. (no change needed)

- **File:** `deploy.sh:L239-L240`
- **Problem:** The new release-consistency gate parses `docker compose config --format json` using `jq`. Because `set -euo pipefail` is active and `jq` is not verified before this line, a production host lacking `jq` will see the script abort after images are pulled but before they are restarted, making the new deployment path fail even though Docker is working.
- **Fix:** Add a preflight dependency check before step 2 (the image pull) so the script fails early with a clear message. For example: `if ! command -v jq >/dev/null 2>&1; then echo "ERROR: jq is required to verify pulled image revisions." >&2; exit 1; fi`

---

#### ~~⚠️ P2 — Server smoke test omits required environment variables~~ ❌ **Not an issue**

> **Note:** The variables are passed. `smoke_env` is a YAML folded scalar that resolves to the single-line string `-e WORKSPACE_MASTER_PASSWORD=ci-smoke-test -e WORKSPACE_BACKEND=filesystem -e WORKSPACE_DATA_DIR=/data/server -e WORKSPACE_BIND_ADDR=127.0.0.1:12349`, and the deliberately unquoted `${{ matrix.smoke_env }}` interpolation is word-split by the shell into individual `-e` flags. Verified empirically rather than by inspection: running the real production image with exactly that expansion, `docker inspect` shows the container receiving all four variables (WORKSPACE_MASTER_PASSWORD included), and it boots, seeds, listens and passes its `nc -z` probe with zero errors. Worth checking, since this workflow only runs on master and a failure here would not surface on a PR. (no change needed)

- **File:** `.github/workflows/publish-images.yml:L46-L53`
- **Problem:** The `smoke_env` for the server sets several environment variables but omits `WORKSPACE_MASTER_PASSWORD` from the actual bash execution string passed in the `docker run` step. If the production server strictly requires this password to boot, the smoke test will fail.
- **Fix:** Ensure the environment variables defined in `smoke_env` are actually passed to the `docker run` command during the smoke test, e.g., by evaluating the string: `docker run -d --name "$name" $(echo "${{ matrix.smoke_env }}") "$IMAGE:$SHA_TAG"`. Alternatively, refactor `smoke_env` into an environment file or standard command-line flags.

---

### ✅ APPROVE

There are no P1 issues and only two P2 issues, meeting the criteria for approval. The PR introduces a much-needed improvement to the deploy process. The `jq` dependency should be addressed before deploying to a host that might lack it, and the smoke test configuration should be verified so the CI workflow doesn't fail.

### Quick Reference
- **P2**: deploy.sh introduces undeclared jq dependency without a preflight check
- **P2**: Server smoke test omits required environment variables

---

PR #62, opened 2026-08-08
fix(deploy): guard the rolling restart on services the compose file declares

Tjemmmic:
## Problem

`deploy.sh` derives its pull list from `docker compose config --services` precisely so `ui` and `internal-service` can be dropped for a server-only deployment — `docker-compose.production.yml` documents that as supported, and the pull-phase comment says so explicitly. The rest of the script did not follow suit.

The rolling restart ran `up -d --no-deps internal-service` and `up -d --no-deps ui` unconditionally, each followed by a `wait_for_port`. An operator who followed our own documentation therefore hit `no such service: ui` and, because of `set -euo pipefail`, aborted the deploy **after the server had already been swapped to its new image** — a half-applied deploy, which is the single outcome the ordering in this script exists to prevent. Guarding the pull but not the restart did not prevent that failure; it deferred it to the most expensive possible moment.

The closing summary had a quieter version of the same bug: it always printed `Local access: http://localhost:8080` and the agent WebSocket URL, so a correctly slimmed-down deployment ended by advertising two endpoints nothing was listening on.

## Change

Adds `service_declared()`, reusing the `PULL_SERVICES` list the pull phase already derives, and applies it to both optional restarts and both summary lines.

`server` is deliberately **not** guarded: the existing check aborts when it is absent, so a guard there would turn "the compose file is malformed" into "nothing was deployed, exit 0".

## Verification

- `bash -n deploy.sh` passes — the check `validate.yml` runs on this file.
- The real `service_declared` definition was extracted from the script and exercised across three topologies: full stack, server-only, and server + internal-service.
- The full derivation chain was run end-to-end against the real `docker-compose.production.yml` (both services restart, both endpoints printed — **behaviour unchanged**) and against a server-only compose file (both skipped, no endpoints advertised).
- The original bug was reproduced rather than assumed: `docker compose -f <server-only.yml> up -d --dry-run --no-deps ui` returns `no such service: ui`, exit 1.

## Known gap

This guard has no behavioral test in CI — `validate.yml` only runs `bash -n` on `deploy.sh`. Covering it properly would mean extracting the service-selection logic into a standalone script, the way `scripts/verify-image-revisions.sh` was extracted for exactly that reason ("an untested safety gate is a liability"). That felt like overreach for this fix, but it is the natural follow-up if this area grows.

PR #63, opened 2026-08-16
fix(deploy): guard restarts on the services the compose file declares

Tjemmmic:
## Overview

Fixes a deploy that breaks when an operator follows our own documentation.

`docker-compose.production.yml` documents that `ui` and `internal-service` are droppable for a server-only deployment. But `deploy.sh` restarted them unconditionally:

```bash
docker compose -f "$COMPOSE_FILE" up -d --no-deps ui
```

An operator who took that option got `no such service: ui` — **after** the server had already been swapped to its new image. A half-applied deploy, caused by following the docs.

## What changed

**Pull, revision-check and restart all derive from one selection.** `scripts/select-deploy-services.sh` decides which of the deployable services the compose file actually declares; `deploy.sh` uses that single set for every phase. Getting it wrong in either direction is a half-applied deploy — too few silently skips a declared service while still reporting success, too many restarts onto an image that was never pulled — so the selection is a standalone, directly tested script rather than inline logic wedged between a pull and a production restart.

**A slimmed deploy that would strand a running service refuses.** `docker compose up` does not remove containers for services the file no longer declares; it warns about orphans and leaves them running. Guarding the restart alone would therefore turn the loud failure above into a silent one — deploy succeeds, old `ui` keeps serving its previous image on `:8080`. Instead the deploy stops **before** pulling or restarting anything, names the containers, and prints the command to clear them.

It deliberately does not remove them itself. Removal has to match containers by the `com.docker.compose.project` label, which spans every checkout and account on the daemon, so doing it safely requires a lock proving no other deploy is mid-flight. That is a large and destructive mechanism for a transition that is rare and takes one command by hand; reporting is race-free by construction. Automatic cleanup is prototyped in #62 if we ever want it.

**A server-only deployment no longer trips the release gate.** `verify-image-revisions.sh` required two images to compare and treated one as a usage error, so the deploy aborted at the gate *after* the pull had succeeded. One image is a valid deployment: it is still inspected — an un-inspectable image is still refused — there is simply nothing to cross-check. Found by the new end-to-end test.

## Testing

`scripts/test-deploy-services.sh` runs the **real** `deploy.sh` end to end against fixture compose files with a recording `docker` stub on `PATH`, asserting what it actually did: which services it pulled and restarted, that it never touches one the file does not declare, and that it exits non-zero when it cannot finish. Seven cases covering every compose shape, both directions of the slim transition, and a `docker compose run` job that must not be mistaken for a stale service.

Each assertion was mutation-tested — the fix reverted, confirming the case fails — because a guard that cannot fail is not a guard. Both scripts are also exercised directly in CI (`validate.yml` → `deploy-gate-tests`) against real images.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -424,6 +424,68 @@ jobs:
           fi
           echo "PASS: mismatched images are refused."
 
+      - name: Deploy service selection covers every compose shape
+        # This selection decides what is pulled, revision-checked AND restarted. Both error
+        # directions are deployment-critical: too few silently skips a declared service while
+        # still reporting success, too many runs `up -d` on an image that was never pulled.
+        # docker-compose.production.yml documents that ui and internal-service are droppable,
+        # so the slimmed shapes are supported configurations, not hypotheticals.
+        run: |
+          set -euo pipefail
+          # Build fixtures from the real production compose so this tracks the actual file.
+          # Dangling depends_on entries must be stripped or compose rejects the project.
+          python3 - <<'PYEOF'
+          import yaml, pathlib
+          base = yaml.safe_load(open('docker-compose.production.yml'))
+          shapes = {'full':['server','internal-service','ui'], 'server-only':['server'],
+                    'server-ui':['server','ui'], 'server-is':['server','internal-service'],
+                    'no-server':['ui']}
+          for name, keep in shapes.items():
+              d = yaml.safe_load(yaml.safe_dump(base))
+              for svc in list(d['services']):
+                  if svc not in keep: d['services'].pop(svc)
+              for svc, cfg in d['services'].items():
+                  dep = cfg.get('depends_on')
+                  if isinstance(dep, dict):
+                      kept = {k: v for k, v in dep.items() if k in keep}
+                      cfg['depends_on'] = kept
+                      if not kept: cfg.pop('depends_on')
+                  elif isinstance(dep, list):
+                      kept = [x for x in dep if x in keep]
+                      cfg['depends_on'] = kept
+                      if not kept: cfg.pop('depends_on')
+              pathlib.Path(f'/tmp/fx-{name}.yml').write_text(yaml.safe_dump(d))
+          PYEOF
+
+          check() {  # <fixture> <expected space-separated services>
+            local got
+            got=$(./scripts/select-deploy-services.sh "/tmp/fx-$1.yml" | tr '\n' ' ' | sed 's/ $//')
+            if [ "$got" != "$2" ]; then
+              echo "::error::selection for '$1' was [$got], expected [$2]. A wrong set here means a half-applied deploy."
+              exit 1
+            fi
+            echo "  $1 -> $got"
+          }
+          check full        "server internal-service ui"
+          check server-only "server"
+          check server-ui   "server ui"
+          check server-is   "server internal-service"
+
+          # A compose file with no server must ABORT, not return an empty set - "nothing to
+          # deploy" reported as a successful no-op deploy is the worst possible outcome.
+          if ./scripts/select-deploy-services.sh /tmp/fx-no-server.yml >/dev/null 2>&1; then
+            echo "::error::a compose file with no 'server' service was accepted; it must abort."
+            exit 1
+          fi
+          echo "  no-server -> correctly aborts"
+
+          # An unreadable file must fail loudly rather than yield an empty selection.
+          if ./scripts/select-deploy-services.sh /tmp/definitely-absent.yml >/dev/null 2>&1; then
+            echo "::error::a missing compose file was accepted; it must abort."
+            exit 1
+          fi
+          echo "  missing file -> correctly aborts"
+
       - name: THREE images are all checked, not just the first two
         # deploy.sh now passes three images (server, internal-service, ui). The script iterates
         # over "$@", so this works - but nothing pinned it, and the failure mode of a hardcoded
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -424,6 +424,28 @@ jobs:
           fi
           echo "PASS: mismatched images are refused."
 
+      - name: A SINGLE image is still inspected
+        # A server-only deployment passes exactly ONE image, so this is a supported production
+        # shape rather than a degenerate call - and it must still be inspected. "Nothing to
+        # compare against" removes the cross-image check, not the provenance check.
+        #
+        # The failure direction is the load-bearing one. An `if [ "$#" -eq 1 ]; then exit 0`
+        # fast path looks harmless and passes both success cases below, but it returns before
+        # `docker image inspect` runs - so a server-only deploy whose pull silently failed, or
+        # whose tag does not exist, would sail through the gate built to catch precisely that.
+        # The end-to-end test cannot see this: its docker stub always inspects successfully.
+        run: |
+          set -euo pipefail
+          ./scripts/verify-image-revisions.sh rev-a
+          echo "PASS: a single labelled image deploys."
+          ./scripts/verify-image-revisions.sh rev-none
+          echo "PASS: a single unlabelled image warns but does not block."
+          if ./scripts/verify-image-revisions.sh definitely-not-pulled:latest 2>/dev/null; then
+            echo "::error::The gate allowed a single image it could not inspect. A server-only deploy whose pull never completed would restart production on a stale or missing image."
+            exit 1
+          fi
+          echo "PASS: a single un-inspectable image is refused."
+
       - name: Deploy restart guards - run deploy.sh end to end
         # The selector test below proves the LIST is right. This proves deploy.sh ACTS on it:
         # it runs the real script against fixture compose files with a recording `docker` stub on
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -468,6 +468,13 @@ jobs:
           # PyYAML is declared rather than inherited from the runner image: a runner change would
           # otherwise turn this deterministic gate into an opaque ModuleNotFoundError.
           python3 -c 'import yaml' 2>/dev/null || pip install --quiet pyyaml
+          # Fixtures are written into the CHECKOUT, not /tmp. `docker compose -f <file>` takes the
+          # file's directory as the project directory, so a fixture in /tmp resolves `env_file:`
+          # paths and ${VAR} references against /tmp instead of the repo root - the test could then
+          # fail, or silently resolve a variable to empty, for reasons having nothing to do with
+          # service selection, and it would not match how deploy.sh invokes the same command.
+          trap 'rm -f fx-*.yml' EXIT
+
           # Build fixtures from the real production compose so this tracks the actual file.
           # Dangling depends_on entries must be stripped or compose rejects the project.
           python3 - <<'PYEOF'
@@ -490,12 +497,12 @@ jobs:
                       kept = [x for x in dep if x in keep]
                       cfg['depends_on'] = kept
                       if not kept: cfg.pop('depends_on')
-              pathlib.Path(f'/tmp/fx-{name}.yml').write_text(yaml.safe_dump(d))
+              pathlib.Path(f'fx-{name}.yml').write_text(yaml.safe_dump(d))
           PYEOF
 
           check() {  # <fixture> <expected space-separated services>
             local got
-            got=$(./scripts/select-deploy-services.sh "/tmp/fx-$1.yml" | tr '\n' ' ' | sed 's/ $//')
+            got=$(./scripts/select-deploy-services.sh "fx-$1.yml" | tr '\n' ' ' | sed 's/ $//')
             if [ "$got" != "$2" ]; then
               echo "::error::selection for '$1' was [$got], expected [$2]. A wrong set here means a half-applied deploy."
               exit 1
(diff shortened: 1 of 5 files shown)


Tjemmmic:

> **Resolution (7a0a2e8):** 1 addressed · 1 acknowledged. 

### Summary

The PR moves deployment service selection into a dedicated, CI-tested selector and drives pull, revision-check and restart from that single list, so the documented slim (server-only) deployments finally work end to end instead of dying mid-restart; it also adds a stale-container preflight and legitimizes single-image verification. The core fix is well-reasoned and well-tested, but the new preflight has two correctness gaps: it misreports stopped containers as running, and its check is not atomic against concurrent deployments.

### Issues Found

#### ~~⚠️ P2 — Stopped containers are flagged as running stale services, blocking valid slim deploys~~ ✅ **Addressed**

> **Note:** Confirmed and fixed in `eba84c3` — the query is now `docker ps -q`. Verified against a real daemon which states it returns: running, paused and restarting are RETURNED; exited and created are hidden — so the refusal now covers exactly the containers that can still serve, and the "still running" wording is accurate. `41986d7` makes the stub honour `-a` the way docker does and adds two cases: a slim deploy proceeds past exited leftovers, and still refuses when one of them is running. Re-adding `-a` fails both.

- **File:** `deploy.sh:L257-L265`
- **Problem:** The preflight uses `docker ps -aq`, which lists stopped and exited containers as well as running ones, yet every returned ID is treated as a service that is still running and serving traffic ('they are still running', 'kept serving their old images'). A harmless exited container left over from an earlier full-stack topology will therefore refuse every slim deployment and instruct the operator to force-remove it, even though nothing is listening. The integration stub's `ps` handler echoes a container ID for any PREEXISTING service without modeling state, so this significant new decision path has no test that would expose the bug.
- **Fix:** Drop the `-a` so only running containers trigger the refusal (`docker ps -q --filter ...`), or inspect each returned container and filter explicitly by running state. Add an integration case where an omitted service has only an exited container and assert deployment proceeds, alongside the existing running-container refusal case.

---

#### ~~⚠️ P2 — Stale-container preflight has a TOCTOU race with concurrent deployments~~ ℹ️ **Acknowledged**

> **Note:** The race is real and the critique of the comment was correct — "race-free by construction" overstated it, and `7a0a2e8` replaces that claim with the bound that actually holds: because the check only ever reports, losing the race costs a missed refusal, never a container removed out from under another deploy and never a half-applied topology this script created; the leftover stays where master already leaves it and the next deploy refuses on it. The lock itself is deliberately out of scope here: host-wide serialization is a change to how deployments are provisioned rather than a fix to this check, and it stalled on an unresolved question (a canonical lock provisioned mid-deploy leaves the two runs contending on different inodes; both known fixes require admin-provisioned lock infrastructure before any deploy can run). Tracked in #64 with the prototype preserved.

- **File:** `deploy.sh:L251-L266`
- **Problem:** The script checks for omitted-service containers once, then performs pulls and several rolling restarts without holding any project-level lock. Another deployment against the same Compose project can create an omitted `ui` or `internal-service` container after this check runs, letting this deployment finish successfully while a service its compose file drops keeps running — the exact half-applied topology the guard exists to prevent. The in-code rationale ('reporting is race-free by construction') covers races during removal, not invalidation of the check itself by a concurrent deploy.
- **Fix:** Acquire a host-wide lock keyed by the resolved Compose project name before the stale-container query and hold it through the complete pull, verification and restart sequence, with cleanup on exit. Every entry point that runs this deployment workflow for the same project must take the same lock.

---

### ✅ APPROVE

> **Status:** both P2s dispositioned — the `docker ps -q` fix landed with coverage; the deployment lock stays out of scope and is tracked in #64.

Both reviewers converge on the same race condition, and the diff confirms it, as well as confirming the `docker ps -aq` stopped-container defect — so the merged review carries two P2s. Neither is blocking: both live in the new refusal path and fail in the safe direction (refuse to deploy rather than half-apply or mutate the running stack), the race requires overlapping deployment activity on the same project, and the stopped-container case has a printed one-command workaround. That is below the changes-requested threshold of any P1 or 3+ P2s, so the PR is approved — but the `docker ps -q` fix should land promptly, since the current behavior will spuriously block documented slim deployments on hosts with leftover exited containers, and the deployment lock should follow to make the topology invariant actually hold.

### Quick Reference
- **P2**: Stopped containers are flagged as running stale services, blocking valid slim deploys
- **P2**: Stale-container preflight has a TOCTOU race with concurrent deployments

---

Tjemmmic:

> **Resolution (512d2fd):** 2 addressed. 

### Summary

The PR derives pull, verification, and restart targets from a single tested selector and adds a stale-service preflight, with genuinely strong test coverage. However, the new guard has two holes that permit exactly the outcome it exists to prevent: an acknowledged but unserialized TOCTOU window across concurrent deploys, and exited `restart: always` containers that are dismissed on an incorrect claim about Docker restart semantics and can resurrect after a successful slim deploy.

### Issues Found
**2 total** — 2 P1 (blocking) · 0 P2 (should fix) · 0 P3 (nice to have)

#### ~~❗ P1 — Stale-service preflight has an unserialized TOCTOU race between concurrent deploys~~ ✅ **Addressed**

> **Note:** Fixed in `512d2fd`. You were right that this belongs in deploy.sh: a `flock` keyed by the already-resolved project name, taken before the stale check and held for the whole run via `exec 9>` so the kernel releases it at exit — no unlock path and no trap to misfire, and the file is never unlinked. The earlier deferral was calibrated to a lock that would have *authorized* `docker rm -f`, where a lock that fails to serialize means destroying another deploy's containers; this one only serializes, so an account that never contends is left exactly where the script already was. Lock path is keyed off `HOME` alone, deliberately not `XDG_RUNTIME_DIR`, which is set in a login session and absent under cron and would hand the same operator two different locks. `flock` is now a checked prerequisite alongside `jq`. Covered by two cases: a deploy refuses while another holds the lock (and refuses before pulling or restarting), and a second sequential deploy re-acquires it, so the lock cannot leak.

- **File:** `deploy.sh:L241-L300`
- **Problem:** The script checks for running containers of dropped services once, then pulls and restarts without holding any lock. A concurrent deploy of the same compose project from a checkout that still declares `ui` or `internal-service` can start one after this check, and this run will pull, restart, and print "Deploy complete" while an undeclared service keeps serving an old or independently selected image — the exact silent mixed-topology deployment the guard was added to prevent. The diff's own comments acknowledge this window and defer the fix, but the guard cannot be considered reliable while the window is open, and the lock does not actually require provisioning-level changes: every deployment goes through deploy.sh, which already resolves the project name (`preflight_project`).
- **Fix:** Acquire a project-scoped lock (e.g., `flock` on a fixed path keyed by the resolved compose project name) before the stale-service check and hold it through the final restart and health checks; exit with a clear message if another deploy holds it. All deploy.sh invocations for that project then serialize without changes outside the script.

---

#### ~~❗ P1 — Exited containers with restart: always are ignored and can resurrect after a successful slim deploy~~ ✅ **Addressed**

> **Note:** Correct, and the comment was wrong — `always` is precisely the policy that leaves a hand-stopped container down until the daemon restarts and then starts it again; that is the only thing separating it from `unless-stopped`, so my claim would have made the two policies identical. Fixed in `1f61335`: the query is `docker ps -aq` again (oneoff=False kept) and each candidate is classified by `docker inspect` on `.State.Status` and `.HostConfig.RestartPolicy.Name`. Refuses on running, paused and restarting — paused is included because the daemon still holds the published ports — and on stopped containers whose policy is `always`, reported as `restarts on reboot` with the reason printed. Proceeds past `no`, `unless-stopped` and `on-failure`. A container that cannot be inspected cannot be classified, so it is refused rather than assumed harmless. Suite is now 14 cases; every one of these branches was mutation-checked.

- **File:** `deploy.sh:L267-L295`
- **Problem:** The preflight deliberately uses `docker ps -q`, excluding exited containers, justified by the comment that an exited container does not come back on its own because under `restart: always` "the daemon restarts it immediately." That claim is wrong for manually stopped containers: Docker suppresses `restart: always` only until the Docker daemon restarts, after which the container starts again. A slim deployment can therefore succeed while retaining an exited, undeclared `ui` or `internal-service` container that resurrects with stale code and ports after a host reboot — the same silent stale-service outcome the guard exists to prevent, masked by a success message. The diff's blanket dismissal also conflicts with its own UX goal, since only `restart: always` exited containers are latent rather than inert.
- **Fix:** Query dropped-service containers with `docker ps -aq` (keeping the oneoff=False filter) and refuse when any candidate's restart policy is `always` — e.g., inspect `HostConfig.RestartPolicy` per candidate — while continuing to proceed past `no`/`unless-stopped` exited containers, which genuinely cannot resurrect. Correct the inaccurate comment and extend `test-deploy-services.sh` so the exited-leftover case models an exited `restart: always` container and expects refusal (the stub already keys off `-a`).

---

> **Status:** both P1s addressed in `1f61335` and `512d2fd`; nothing deferred.

Both P1s allow a deployment to report success while a stale, undeclared service is serving or will serve after a daemon restart — precisely the failure mode this PR's preflight was added to eliminate, so shipping the guard with these holes open gives a false sense of safety. The concurrency fix is small and belongs in deploy.sh (which already resolves the project name), not in deferred provisioning work, and the exited-container dismissal rests on a factually incorrect reading of Docker's `restart: always` semantics that at minimum must be corrected with a policy-aware refusal. The selector extraction, single-image verification fix, and test suite are excellent and should be kept; only the two guard holes block merge.

### Quick Reference
- **P1**: Stale-service preflight has an unserialized TOCTOU race between concurrent deploys
- **P1**: Exited containers with restart: always are ignored and can resurrect after a successful slim deploy

---

Tjemmmic:

> **Resolution (e7b6d9c):** 2 addressed · 1 acknowledged. 

### Summary

The PR correctly derives the deploy set from the compose file, refuses to deploy over leftovers of dropped services, and adds a project lock, with unusually thorough test coverage. However, the lock is acquired too late (after the git-update step and after service selection) to fully serialize runs from the same checkout, and the stale-container recheck conflates a failed Docker query with 'container gone', failing open — either defect can reproduce the silent half-applied deployment this PR exists to prevent.

### Issues Found
**3 total** — 2 P1 (blocking) · 1 P2 (should fix) · 0 P3 (nice to have)

#### ~~❗ P1 — Deployment lock is taken after the git update and service selection, so a concurrent run can still mutate the checkout mid-deploy~~ ✅ **Addressed**

> **Note:** Confirmed — `git pull` is at deploy.sh:175 and the project lock was at :298, so the chicken-and-egg you describe is real: the project name is only knowable after reading the compose file. Fixed in `e42635b` with a second lock taken before anything reads or writes the checkout, as you suggested. Rather than hashing the path, it locks the checkout **directory's own inode** (`exec 8<"$PWD"`): the identity is then exact, with no name collisions between checkouts and no path-length limit, and two runs from the same directory necessarily open the same inode. Verified against a real daemon-free run — with a holder, it refuses before step 1; without, it proceeds. Both locks are non-blocking and taken in a fixed order (checkout, then project), so they cannot deadlock. Covered by a case that holds the checkout lock and asserts the second run refuses **before `[1/4]`**; moving the lock after step 1 fails that assertion.

- **File:** `deploy.sh:L222-L322`
- **Problem:** The lock is keyed on the compose project name, which is only derivable by reading the compose file, so it is acquired in step 2 — after step 1's `git pull` and after the service selection has already read the compose file. A second invocation from the same checkout executes its `git pull` while the first holds the lock, and can change docker-compose.production.yml (or deploy.sh / scripts) between the first run's selection and its later `config --format json` and `up -d` reads. The winner then pulls and verifies against one revision of the file while restarting against another, recreating the interleaved half-applied deployment the lock was added to prevent. (The second run is refused at flock before touching Docker state, but its git step has already mutated the shared checkout.)
- **Fix:** Serialize the whole run: take a bootstrap lock before any git operation, keyed on something stable before mutable state is read (e.g. the checkout path or a configured deployment identifier), or an account-wide lock — then perform git update, selection, preflight, stale check, pull, verify and restart entirely while holding it (the project-scoped lock can remain for same-account cross-checkout serialization). Add an integration test where a competing invocation reaches the pre-lock phase and mutates the compose definition.

---

#### ~~❗ P1 — Stale-container recheck treats a failed `docker ps` as proof the container vanished~~ ✅ **Addressed**

> **Note:** Correct, and reproduced: `printf '' | grep -q .` is false, so a failed query took the same branch as "container is gone" — fail-open on the one path whose comment says it fails closed. Worse than incidental, since whatever stopped `docker inspect` answering will usually stop this query too, making the failure the likely case rather than an exotic one. Fixed in `273426b` exactly as you proposed: capture the query, refuse if it fails, and continue only when it succeeded and reported the ID absent. The stub can now fail a `ps --filter id=` query, and a case asserts the deploy refuses and never reaches pull or restart.

- **File:** `deploy.sh:L333-L341`
- **Problem:** When `docker inspect` fails, the code runs `docker ps -aq --filter "id=${cid}" 2>/dev/null | grep -q .` and `continue`s when the condition is false. That condition is false both when the container is gone AND when the `docker ps` query itself failed (stderr is discarded, and with pipefail either leg fails the pipeline). A transient daemon/API failure therefore makes the stated rule — 'Anything still there that we cannot read, we refuse on' — fail open: an unclassifiable dropped-service container is silently skipped and the deploy proceeds over the exact silent half-applied topology the guard exists to stop.
- **Fix:** Capture and validate the query separately: `if ! remaining=$(docker ps -aq --filter "id=${cid}" 2>/dev/null); then <refuse, exit 1>; fi; [ -n "$remaining" ] || continue` — refuse when the exact ID is present, and continue only when the query succeeded and the ID is absent. Extend the docker stub in scripts/test-deploy-services.sh with a failing `ps --filter id=` case.

---

#### ~~⚠️ P2 — Lock scoped to $HOME does not serialize the same compose project across Docker accounts~~ ℹ️ **Acknowledged**

> **Note:** Accurate, and the fair part of the criticism was the wording, which `e7b6d9c` fixes: the comment asserted the snapshot "stays true for as long as anything depends on it" without naming the limit. It now states plainly that the lock lives under HOME while the containers and project labels it protects live on the shared daemon, so two accounts deploying the same project take different lock files and do not serialize at all. The fix you describe — a provisioned host-wide lock with ownership and permissions preventing replacement — is a deployment-provisioning change rather than a script change, and is tracked in #64 along with the container-removal work it also gates. It stays non-blocking for the reason you give: nothing acts on this lock's authority, so an unshared lock costs a missed refusal, never a destructive action.

- **File:** `deploy.sh:L275-L322`
- **Problem:** The lock lives under `$HOME/.cache/citadel-deploy` keyed by project name, but the containers and project labels it protects are scoped to the shared Docker daemon — and the stale check itself matches by project label, which reaches every checkout and account on the daemon. Two accounts with Docker access therefore acquire different lock files for the same project and can interleave pulls, stale snapshots and restarts, so the script's claim that 'the snapshot stays true for as long as anything depends on it' does not hold cross-account. The comments acknowledge the gap and correctly note the lock authorizes nothing destructive (worst case is a missed refusal, i.e. pre-PR behavior), which keeps this below blocking, but the advertised project-level serialization is not actually enforced.
- **Fix:** Either place the project lock in a provisioned host-wide directory shared by every authorized deployer, keyed by Docker daemon/context and the validated compose project name, with ownership and permissions that prevent lock-file replacement; or explicitly enforce and document single-account deployment so the unserialized cross-account case cannot arise silently.

---

> **Status:** both P1s addressed in `273426b` and `e42635b`; the P2 is a documented scope limit, corrected in `e7b6d9c` and tracked in #64.

Two P1 defects defeat the PR's own invariant — a deployment must never report success over a topology it did not verify. The lock is acquired only after the git-update step and service selection, so a concurrent run from the same checkout can still change the compose file between selection and restart; and the stale-container recheck fails open when Docker cannot answer, letting an unclassifiable leftover through. Either can yield a silent half-applied production deploy, so changes are requested despite the otherwise sound selection logic, refusal-before-mutation ordering, and excellent test coverage.

### Quick Reference
- **P1**: Deployment lock is taken after the git update and service selection, so a concurrent run can still mutate the checkout mid-deploy
- **P1**: Stale-container recheck treats a failed `docker ps` as proof the container vanished
- **P2**: Lock scoped to $HOME does not serialize the same compose project across Docker accounts

---

Tjemmmic:

> **Resolution (e7b6d9c):** 1 acknowledged. 

### Summary

This PR is a well-executed safety overhaul: a single derived service selection drives pull, verification and restart; a stale-container preflight refuses slim deploys over live or reboot-resurrectable leftovers; and dual flock serialization (checkout directory plus compose project) covers the two interleaving windows. The reviews flag two concerns: one about the unless-stopped classification (incorrect, contradicted by Docker's documented restart semantics) and one about the per-account lock scoping (real but explicitly acknowledged and non-destructive in the diff).

### Issues Found
**1 total** — 0 P1 (blocking) · 1 P2 (should fix) · 0 P3 (nice to have)

#### ~~⚠️ P2 — Per-account project lock does not serialize cross-account deploys on a shared Docker daemon~~ ℹ️ **Acknowledged**

> **Note:** Accurate, and it matches what deploy.sh already states at the lock: the file lives under HOME while the containers and project labels it protects live on the shared daemon, so two accounts take different locks and do not serialize. Both proposed fixes — an administrator-provisioned host-wide lock keyed by daemon plus project, or a single enforced deployment identity — are deployment-provisioning changes rather than script changes, so neither belongs in this PR; they are tracked in #64, which has been updated to record that this per-account serialization lock has now shipped and that the same provisioning work closes its cross-account gap. Taking it no further here for the reason you give: the lock authorizes nothing destructive, so the worst case is a missed refusal in a scenario that had no protection at all before this change. On the runbook suggestion — there is no operator runbook for deploy.sh in this repo (`docs/PRODUCTION_DEPLOYMENT.md` is an analysis document and does not mention the script), so the limitation is recorded in the code at the lock and in #64 rather than in a doc invented for the purpose.

- **File:** `deploy.sh:L312-L362`
- **Problem:** The lock file lives under ${HOME}/.cache/citadel-deploy, but the stale-container preflight matches containers by com.docker.compose.project label across the entire shared daemon. Two different accounts deploying the same project therefore take different lock files and can run concurrently: one deploy can pass the stale check, after which the other starts a dropped service's container, and the first then reports 'Deploy complete' over a topology it never verified. The diff explicitly documents this scoping and argues it stays non-blocking because the lock authorizes no destructive action — the cost is a missed refusal in a narrow multi-account scenario that was entirely unprotected before this PR, not a destructive action. That reasoning holds for P2 rather than P1.
- **Fix:** Provision a host-wide lock keyed by Docker endpoint plus compose project, with administrator-set ownership and permissions that all deployment accounts can open but none can replace. Alternatively, require all deploys for a daemon to run through one dedicated service account and validate that identity before proceeding. Either closes the cross-account window; document the interim limitation in the deploy runbook until then.

---

### ✅ APPROVE

> **Status:** the single P2 is a documented scope limit, unchanged in this PR and tracked in #64; no code changes were made for this review.

Reviewer B's P1 on unless-stopped is discarded as incorrect: Docker's documented behavior is that a stopped unless-stopped container is not restarted 'even after the Docker daemon restarts', and an exited unless-stopped container observed while the daemon runs must have been hand-stopped (any exit triggers an automatic restart, and an active crash loop reports 'restarting', which the script already blocks). The diff's only-always classification is therefore correct, and the slim-exited test that codifies it is fine. The lock-scoping race from B (P2) and C (P1) is a real, duplicated finding, consolidated here: it is explicitly acknowledged in the diff, gates nothing destructive (worst case is a missed refusal in a rare cross-account scenario, the pre-PR status quo), and is strictly an improvement over the previously lock-free script — so P2, not blocking. With no P1s and only one P2, the change is safe to merge with the lock-provisioning follow-up tracked.

### Quick Reference
- **P2**: Per-account project lock does not serialize cross-account deploys on a shared Docker daemon

---

PR #65, opened 2026-08-24
Cleanup, harden, and complete the product lifecycle

tbraun96:
## Summary

Resets onto Tjemmmic's `master` and takes the app through the whole product
lifecycle — first run, deployment, upgrade, everyday use — fixing what was
broken and putting gates under what was merely asserted.

This is one PR by request. Matching branches exist in both submodules
(`cleanup-and-harden`).

## Product work

- **Workspace theming.** A strongly-typed `WorkspaceTheme` stored in workspace
  metadata, nine presets (Avarok Purple is the default and a literal
  transcription of `index.css`, asserted by test), a live editor built on the
  workspace preview itself, colour wheel with hex and native picker, per-member
  light/dark that stays the member's own choice. New `themes` permission,
  server-side, gated separately from the master password — changing a colour must
  not need the credential that also deletes the workspace.
- **The five product decisions** — revfs tree propagation, file-transfer
  receiver, message reply/edit/delete, ad-hoc group conversations — built and
  shipped.
- **PWA.** Installable, offline-capable, with a real upgrade prompt.

## Defects found and fixed

- **Saving a theme made the app unusable.** `workspace.metadata` is one JSON
  document several features share; initialisation writes `{"initialized": true}`
  there. The handler assigned over it, so a configured workspace reopened its
  setup modal over a working UI, and that modal's backdrop swallowed every click.
  The app rendered perfectly and responded to nothing.
- **The permissions gate was dead for logged-in users.** `getCurrentUserId()`
  returned `connectionInfo?.username`, null unless you had just registered, so
  every permission check silently failed. Two sites: one bailed, one discarded
  valid responses.
- **`node_mutex` bypass** in three read-modify-write paths over `get_all_nodes()`.
- **Dark `destructive` failed WCAG AA** at 3.78:1 — the colour every destructive
  button uses. The preset AA test only checked three pairs, which is how it
  shipped; it now covers every text-carrying fill across both palettes of all
  nine presets, and immediately caught two more in Solarized.
- **No dialog bounded its height**, so a tall one pushed its own footer
  off-screen. On a phone the theme editor's Save and Cancel were unreachable.
- **The theme editor had a critical a11y violation** — `Tabs` with no
  `TabsContent`, emitting `aria-controls` at an element that never existed.

## Testing

- 787 unit tests (from ~150), 59 Playwright specs, all green
- Legacy suite: 48/48 CI-gated, up from 31
- New gates: axe accessibility, Lighthouse baselines, PWA installability, **PWA
  offline behaviour**, bundle budget, 250-line file limit
- ~697 hardcoded sleeps audited; `isVisible()` false-passes removed

Several helpers reported success without checking — eight call sites ignored
`waitForWorkspaceLoaded`'s return value, which hid that it could not recognise a
loaded workspace at phone widths at all. Those now assert.

## Risk

The base `DialogContent` change affects all 27 dialogs; the full Playwright
suite passes. The `citadel-workspace-types` change is protocol-level and needs
both submodules moved together.


citadel-workspace-types/src/structs.rs
Tjemmmic, line 1: [P1] Preserve deserialization of stored permission records

`DomainPermissions` is embedded in persisted `DomainNode` values, and backend reads use `serde_json::from_slice`. Existing records predate `themes`; without a Serde default, loading them fails with `missing field themes` and can make the stored node map unreadable after an upgrade.

Please add a legacy-JSON regression test and default the new field:

```rust
#[serde(default)]
pub themes: bool,
```

package-lock.json
Tjemmmic, line 1: [P1] Regenerate the lockfile from a clean, portable install

This lockfile cannot reproduce a clean CI checkout. Re-running the workflow’s npm major (npm 10.9.4) at this PR head makes `npm ci --dry-run` fail with `Missing: fsevents@2.3.3 from lock file`. npm 12 gets past validation, but Vite then fails because `@esbuild/linux-x64` is declared by esbuild yet has no package record.

Please regenerate the lockfile from a clean install and verify `npm ci` plus the UI build using the workflow’s Node/npm toolchain.

citadel-workspace-server-kernel/src/kernel/command_processor/async_process_command.rs
```diff
@@ -128,6 +128,91 @@ pub async fn process_command_with_user_and_cid<R: Ratchet + Send + Sync + 'stati
             }
         }
 
+        WorkspaceProtocolRequest::UpdateWorkspaceTheme {
+            workspace_id,
```
Tjemmmic, line 131: [P2] Serialize this whole-record theme update

This is an unlocked `get_workspace` → modify → `insert_workspace` read/modify/write. A concurrent workspace/member/theme mutation can read the same old record, and the last writer discards the other update.

Existing workspace mutators also update the denormalized `Domain::Workspace` copy, while this handler writes only the workspace record. Please serialize mutations for the same workspace and preserve that dual-write invariant.

citadel-workspace-server-kernel/src/kernel/command_processor/async_process_command.rs
```diff
@@ -128,6 +128,91 @@ pub async fn process_command_with_user_and_cid<R: Ratchet + Send + Sync + 'stati
...
+            theme,
+        } => {
+            use citadel_workspace_types::structs::Permission;
+
+            let target_id = workspace_id.as_deref().unwrap_or(crate::WORKSPACE_ROOT_ID);
```
Tjemmmic, line 136: [P1] Pass the active workspace ID when saving its theme

Inside this submodule, `WorkspaceAppearanceSection` checks `Themes` permission for the active `workspaceId`, but `handleSave` calls `updateWorkspaceTheme(serializeTheme(next))` without that ID. The server treats an absent ID as `WORKSPACE_ROOT_ID`.

A secondary-workspace owner can therefore see the editor but saving either fails against root permissions or updates the root theme instead. The call should pass `workspaceId`, and the callback should depend on it:

```typescript
await WorkspaceService.updateWorkspaceTheme(serializeTheme(next), workspaceId);
```

sync-wasm-clients.sh
```diff
@@ -60,62 +60,19 @@ DEST3="$WORKSPACE_ROOT/citadel-workspace-client-ts/"
...
 
 # Step 0: Clean
 print_status "Cleaning previous build artifacts..."
 rm -f "$DEST1"/*.wasm "$DEST1"/*.d.ts "$DEST1"/*.js 2>/dev/null || true
 rm -rf "$DEST1/dist" 2>/dev/null || true
 rm -rf "$DEST1/node_modules" 2>/dev/null || true
-# Explicitly remove old package.json to ensure we write fresh
-rm -f "$DEST1/package.json" 2>/dev/null || true
+# NOTE: $DEST1/package.json is tracked in git and is the single source of truth for this
+# package's name, exports, scripts and dependencies. It is deliberately NOT deleted or
```
Tjemmmic, line 87: [P2] Add the TypeScript client tests this script now claims to run

The submodule changes `npm test` from an explicit placeholder to `tsc && node --test "dist/**/*.test.js"`, but no `.test.*` or `.spec.*` sources are committed; the former file is now `manual-smoke.ts`. Running the command at this head exits successfully with 0 tests and 0 suites.

Please add the intended automated tests and make the command fail when no tests are discovered.

Tjemmmic (changes requested):
Requesting changes for five reproducible correctness and build issues. Each finding is documented in a separate review comment.

PR #66, opened 2026-08-24
Audio and video calling, over the Citadel UDP channel

tbraun96:
## Summary

Adds audio and video calling — proven working end to end — and upgrades
Citadel-Protocol to master (`56650e9b` → `a28a3c7c`), which is where the media
transport this builds on landed.

Branches off `cleanup-and-harden` (PR #65) and merges back into it.

## Proven, not asserted

`call-audio-video.spec.ts` passes **8/8**. The load-bearing assertion is a
**remote** participant tile reporting a non-zero `videoWidth`, in both
directions — only possible if a frame was captured, encoded by WebCodecs,
fragmented, carried over the peer's UDP channel through the internal service,
reassembled, decoded and painted.

An earlier version of that assertion matched *every* participant tile. The
self-tile always has video, so it passed **without a single frame crossing the
network**. Excluding the self-tile is what revealed that nothing was pumping
capture into the encoders at all.

`call-group.spec.ts` passes **11/11** with three real participants. Here the
load-bearing assertion is **two DISTINCT remote tiles decoding on each
invitee** — one of those senders is the *other invitee*, so a single-tile
assertion would pass under hub-and-spoke while the mesh was not forming at all.
Both invitees are checked, and so is the caller's fan-in.

The same trap in a different disguise: the caller sees and hears everyone
whether or not the mesh exists, so the seat most likely to be tested is the one
seat that cannot detect the failure.

## Architecture

- **Media rides the peer's UDP channel, not the reliable one.** On a reliable
  ordered channel there is no such thing as a lost packet — congestion becomes
  unbounded latency instead. A call three seconds behind is worse than one that
  dropped a frame.
- **Encoding happens in the browser** (WebCodecs; Opus + AV1/VP9/H.264). It is
  the only route to hardware encoders, and this WASM build has neither threads
  nor SIMD.
- **Signalling rides the reliable path.** Losing a frame costs a sixtieth of a
  second; losing a "call ended" strands someone in a call that is over.
- **Liveness**: presence is announced every 5s and silence past 20s means gone.
  Absence of media frames is deliberately NOT used — a muted participant with
  their camera off sends nothing and is still there.
- Framing, jitter buffering and gap reporting come from the upstream
  `citadel_media` crate rather than being hand-rolled.

## Surfaces

Calling works from every conversation in the app: direct messages, ad-hoc peer
group chats, and office/room chat channels. Group calls use the room roster,
refuse with a stated reason when a room outgrows what a mesh can carry, and
offer "Join call" with a live participant count when one is already running.

Sound is **synthesized** with the Web Audio API — no assets, no licensing, and
autoplay refusal degrades to a silent but fully working call. Waiting and
connecting states replace the bare spinner. Every animation is behind
`motion-safe:`, so reduced-motion gets a static state rather than a slow one.

## Defects found and fixed

Most were invisible from the UI — a call that connects, renders and ticks while
carrying nothing:

- The video encoder **died on creation on both sides** — zero frames encoded.
- **Remote audio was decoded into a track attached to nothing.** Every call was
  silent.
- A **duplicated invite made the callee decline its own call** ~300ms in.
- **`MediaSessionOpened/Failed` had no listener**, so every service-side failure
  was invisible.
- The **UDP channel was consumed rather than borrowed**, so the *second* call
  between two peers could never carry media.
- **`build.rs` overwrote a tracked `package.json`** on every `cargo check`,
  which could wipe the WASM binary and make every service call silently no-op.
  This presented as unrelated UI failures with the service logging nothing.
- The **WASM sync shadowed the root Playwright install**, breaking the test
  suite after every clean deploy with an error that named the wrong problem.

## Group calling was broken in eight places

It had never been run end to end, and every seam was wired from one side only:

- **`startGroupResponseService()` was only called from a retry modal's
  `onRetry`.** In a normal session no group response was ever translated, so the
  sidebar list could never populate.
- **`GroupCreate` sent `bigint[]` where the wire wants `Vec<UserIdentifier>`.**
  The request failed serde inside the client and never left the browser —
  invisible to `tsc`, and only reachable via a create *with* members, which is
  the one thing the dialog requires.
- **The member-state mapper matched wire variants that do not exist** and read
  the recipient's cid as the member.
- **Every invitation was dropped** because the wire carries no inviter username
  and a nameless invite is rejected. Usernames now resolve from the peer
  registry.
- **Auto-accept was local-only**, so the creator's roster never filled.
- **Rust: no invitee could answer an invitation.** `respond_request` demanded a
  peer remote to the inviter, but invitees are normally the P2P *acceptor* and
  an acceptor-only connection carries none. `AcceptMembership` is a
  client-to-server command — the server owns membership — so it now goes on the
  node remote, as `create.rs` already did.
- **Group state lived per hook instance**, bridged by localStorage persistence
  that had never once worked: `JSON.stringify` throws on bigint CIDs and the
  failure was caught and logged. Opening a group bounced to "Group not found".
  Now one module-level store; the dead persistence was removed rather than given
  a bigint-JSON workaround, per the CID policy.
- **The invite carried the caller's roster and the receiver dropped it** — the
  mesh defect above.

Plus one that only appears under fan-out: sending invites with `Promise.all`
reproducibly **lost exactly one of two sends** inside the messenger's async
path, a different victim each run. Signal sends are now serialised — and
bounded, because nothing below the queue has a timeout, so an unbounded queue
would have let one stalled send hold the hang-up behind it.

## Testing

- 1135 unit tests
- Calling **8/8**, group calling **11/11**, accessibility **10/10**, login 5/5
- Legacy suites green; clippy `-D warnings` clean in both Rust workspaces
- 250-line file cap back to **zero violations** (15 files split along
  responsibility seams; tests excluded from the cap, with reasoning)
- Verified from a **clean `docker compose down` + `up`**: all services healthy,
  sync exits 0, no manual cleanup, calling passes immediately
- Lighthouse all four baselines; PWA installability (with app shortcuts) and
  offline behaviour gated; landing bundle 286.7 KB of 300 KB with the media
  pipeline lazily loaded

## Known gaps

- ~~Multi-party media is not proven end to end.~~ **Now proven** — three
  participants, mesh verified from both invitees and the caller.
- Whether a fresh workspace should make everyone in it callable remains an open
  **product** decision. Membership is explicit and admin-driven — verified in
  the server: a user joins `workspace.members` only via the master-password
  path, so a default room has nobody to call and correctly says so. Changing
  that has access-control consequences, so it is surfaced rather than guessed.
- The server broadcasts `EnteredGroup` only to members present at join time, so
  an invitee who joined first never learns of later joiners in its local roster.
  Calls are unaffected — the invite carries the caller's full roster.
- Group state is session-lifetime only; durable state belongs in IndexedDB.
- Cross-machine codec negotiation is covered by unit tests and symmetric E2E
  only; heterogeneous hardware was not available.


citadel-workspace-internal-service/build.rs
```diff
@@ -171,8 +185,19 @@ fn copy_wasm_files(src: &Path, dst: &Path) -> std::io::Result<()> {
   ]
 }"#;
 
-    // Only write package.json to wasm-client-ts, not to public/wasm
-    if dst.ends_with("wasm-client-ts") || dst.ends_with("typescript-client") {
+    // Generated destinations only. typescript-client/package.json is TRACKED IN
```
Tjemmmic, line 188: [P1] Commit a complete generated TypeScript client

The exact CI build (`npm run build` in `typescript-client`) fails at this submodule commit with roughly 50 TypeScript errors. Generation removed required protocol-type imports such as `MessageGroupKey`, `ConnectMode`, `UdpMode`, and `SessionSecuritySettings`, and `InternalServiceResponse.ts` imports five media modules that are not committed.

Please fix the generator/export configuration and commit a complete regenerated tree; patching individual generated files would leave the next regeneration broken again.

citadel-workspace-client-ts/src/WorkspaceClient.ts
```diff
@@ -313,6 +322,35 @@ export class WorkspaceClient extends InternalServiceWasmClient {
...
+  sendMediaFrame(
+    localCid: bigint,
+    peerCid: bigint,
+    track: number,
+    kind: number,
+    timestamp: number,
+    flags: number,
+    payload: Uint8Array,
+  ): void {
+    this.getWorkspaceWasmModule().send_media_frame(
```
Tjemmmic, line 402: [P1] Route call media through the leader/follower architecture

Follower tabs intentionally finish initialization with `service.client = null`, but `WebSocketCallTransport` accesses that client directly: `openSession()` throws `Not connected`, `closeSession()` returns, and `sendFrame()` silently drops every frame.

Calls started or answered outside the elected leader tab therefore cannot establish media. Open/close need the proxy-aware request path, and the high-rate media path needs a deliberate follower-to-leader route (or the UI must explicitly restrict calls to the leader tab).

citadel-workspace-client-ts/pkg/citadel_internal_service_wasm_client.d.ts
```diff
@@ -31,6 +31,21 @@ export function restart(ws_url: string): Promise<void>;
...
+ * the hot path: frames arrive 30-60 times a second per track, and routing each
+ * through a JsValue and serde-wasm-bindgen would deserialize a whole request
+ * object — payload included — per frame. This takes the bytes directly and
+ * builds the request in Rust.
+ *
+ * Fire-and-forget by design. There is no ack to wait for; a frame that cannot
+ * be queued is a frame worth dropping, because by the time a retry arrived it
+ * would be too late to play.
+ */
+export function send_media_frame(local_cid_str: string, peer_cid_str: string, track: number, kind: number, timestamp: number, flags: number, payload: Uint8Array): void;
```
Tjemmmic, line 47: [P1] Enforce media-session ownership on send and close

`MediaOpen` uses the localhost connection UUID to decide whether an existing session belongs to the requester, but `handle_send` and `handle_close` do not verify `session.owner() == uuid`.

After a reconnect replaces a session, delayed requests from the old localhost connection can inject frames into or terminate the new session. Please enforce ownership before exposing the outbound handle or taking/closing the session, and cover delayed old-owner requests in tests.

ARCHITECTURE.md
```diff
@@ -303,6 +303,104 @@ const username = message.PeerRegisterNotification.username;  // undefined!
...
+encodes once and fans the same bitstream to everyone, and picking per-peer would
+mean an encoder instance per participant.
+
+### Frame path
+
+```
+capture (getUserMedia)
+   → CapturePump            MediaStreamTrackProcessor, or a canvas fallback
+   → VideoEncoder/AudioEncoder   WebCodecs, latencyMode 'realtime'
+   → send_media_frame        WASM binding, synchronous and unawaited
```
Tjemmmic, line 355: [P1] Bound the high-rate media queues

The outbound fragment sink calls `OutboundUdpSender::unbounded_send`, and inbound frames are placed on an `UnboundedSender<InternalServiceResponse>`. These sends fail only after the consumer closes, not when it is congested, so a slow UDP path or browser WebSocket accumulates stale media and memory instead of dropping frames.

The sender-side `applyQualityReport` method still has no production caller. Please separate lossy media from reliable control traffic and use a bounded/latest-frame queue with observable congestion feedback.

citadel-workspace-client-ts/src/WorkspaceClient.ts
```diff
@@ -313,6 +322,35 @@ export class WorkspaceClient extends InternalServiceWasmClient {
...
+   * Send one encoded media frame to a peer.
+   *
+   * Synchronous and unawaited on purpose. Frames arrive 30-60 times a second
+   * per track; a promise per frame would allocate thousands of microtasks a
+   * minute for a result nobody inspects, and there is nothing to await — a
+   * frame that cannot be queued is one worth dropping, because a retry would
+   * arrive too late to play.
+   */
+  sendMediaFrame(
+    localCid: bigint,
```
Tjemmmic, line 394: [P1] Tear down the call runtime when the local CID changes

`CallLayer` remains mounted across login and workspace routes and polls the current CID, but `useCallRuntime.ensureManager()` returns an existing `managerRef` before considering the new `selfCid`. An in-flight manager created for the old CID can also finish after identity changes.

After logout, reconnect, or account switching, calling can remain bound to the previous CID. Please end and rebuild the runtime when the CID changes and invalidate construction started for an older identity.

Tjemmmic (changes requested):
Requesting changes for five reproducible build and runtime reliability issues. Each finding is documented in a separate review comment.

PR #91, opened 2026-09-04
feat(ui-image): publish a loopback agent origin, from one value, into the CSP and the page

tbraun96:
A hosted page cannot reach the agent on a visitor's machine: it derives its
socket as same-origin /ws and the public stack keeps that proxy off, because
one shared agent would hold every user's ratchet keys. The way through is the
one certified.sh uses: a name the operator points at 127.0.0.1 and holds a
certificate for, so the page dials wss://local.example.com:12345 and lands on
the visitor's own loopback.

LOOPBACK_AGENT_ORIGIN is the fourth runtime variable the UI image renders.
16-validate-runtime-vars.sh admits only a bare wss://host:port -- the value is
substituted into a quoted nginx string and an HTML attribute, so a quote or a
semicolon is an injection, not a typo -- and the template writes it into the
CSP's connect-src and into <meta name="citadel-loopback-agent"> from the same
variable, so what the app dials and what the policy permits cannot disagree.

The CSP was six byte-identical copies, one per location, because add_header
does not inherit; it is now one map that every location references.

provision-tenant.sh --loopback-host NAME writes the origin (needs --topology
full: only a served UI dials a visitor's agent); without the flag the key is
present and empty, so the compose default never applies silently.

Guards: scripts/test-ui-runtime-config.sh (18 assertions, no image needed;
runs in CI) -- validator accepts/refuses, CSP defined once, six references,
origin in connect-src and meta, empty leaves both alone. Controls: validator
check removed admits every breakout; a seventh literal CSP fails; the
sub_filter removed fails. test-provision-tenant.sh: four new cases plus the
.env contents, control on the topology refusal.

Pairs with citadel-workspace-ui #23, which reads the meta.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -220,6 +220,11 @@ jobs:
 
       - name: Tenant provisioning refuses what it must
         run: bash scripts/test-provision-tenant.sh
+      - name: UI runtime config admits only what nginx can safely render
+        # The validator that runs before nginx renders, and the rendered template: the CSP is
+        # one map, every location references it, and LOOPBACK_AGENT_ORIGIN reaches the policy
+        # and the page meta from the same value. Needs envsubst (gettext-base, on the runner).
+        run: bash scripts/test-ui-runtime-config.sh
 
 
       - name: Every administrator promotion sits behind a gate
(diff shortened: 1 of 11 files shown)


tbraun96:
Dependency note: this PR's image smoke (`smoke-ui-ws.sh` section 5) asserts the served `index.html` carries `<meta name="citadel-loopback-agent">`, which comes from citadel-workspace-ui #23. The parent CI builds the UI image from the pinned submodule commit, so that assertion is expected to fail here until #23 merges and this branch bumps the `citadel-workspaces` pointer to include it (a prepared bump for #19/#20 exists locally and will be extended). Order: UI #23 → pointer bump on this branch → green.

tbraun96:

PR #92, opened 2026-09-05
fix(release): the agent people download could not be reached by a browser

tbraun96:
release-agent.yml built the submodule's citadel_service_bin, which starts the
kernel on the raw TCP interface. The docker image runs this repo's
citadel-workspace-internal-service, which wraps the kernel in the WebSocket
interface and the Origin allowlist. agent-v0.1.0 shipped the TCP one: it
bound its port, passed every smoke assertion (executable, refuses without
--bind, right architecture, listens) and closed the connection on the first
WebSocket handshake. Measured against the published macos-arm64 asset: the
handshake gets '<connection closed>'; the docker agent answers 101.

The workflow now builds and packages citadel-workspace-internal-service,
with a `vendored` feature forwarded to the kernel crate so OpenSSL is
static, as before. smoke-agent.sh performs a real handshake: the allowed
origin must be answered 101 and a foreign origin refused 403, which proves
the allowlist is in the shipped binary. Its curl pipeline had to tolerate a
closed connection -- under pipefail + set -e the control died before the
assertion could name the failure. Control: the v0.1.0 archive now fails at
the handshake; a package of the WebSocket binary passes.

docs/AGENT_README.md gains --allowed-origins, which this binary requires
and the README's bare invocation would have tripped on.

A release build with `--features vendored` is running locally; result to follow in a comment.

### DIFF
--- a/.github/workflows/release-agent.yml
+++ b/.github/workflows/release-agent.yml
@@ -87,9 +87,14 @@ jobs:
       # DOWNLOADED and run on someone else's machine, so linking against
       # whatever OpenSSL the build agent happened to have is a runtime failure
       # waiting to happen on a machine that has a different one, or none.
+      # THE BINARY THE BROWSER CAN TALK TO. citadel-workspace-internal-service (this repo)
+      # wraps the kernel in the WebSocket interface and the Origin allowlist; it is what the
+      # docker image runs. The submodule's citadel_service_bin starts the kernel on the raw
+      # TCP interface -- a browser cannot open that -- and agent-v0.1.0 shipped it: the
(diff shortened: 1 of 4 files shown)


tbraun96:
Local proof on macOS arm64: `cargo build --release -p citadel-workspace-internal-service --bin citadel-workspace-internal-service --features vendored` succeeds; the binary has no dynamic libssl/libcrypto references (OpenSSL is embedded); packaged as the workflow will package it, `scripts/smoke-agent.sh` reports `101 Switching Protocols` for the allowed origin and `403 Forbidden` for a foreign one. The published agent-v0.1.0 archive fails the same script at the handshake with `<connection closed>`.

PR #94, opened 2026-09-05
feat(provision): the vhost serves the loopback certificate, and loads on the nginx it is for

tbraun96:
A tenant with --loopback-host has a certificate for that name (acme.sh,
installed under <tenant>/loopback/). Each user's agent fetches it at start
from https://<domain>/agent/loopback.pem and .key, so the vhost now serves
exactly those two files from that directory, uncached -- a renewed
certificate must reach the next agent start. No listing, nothing else in the
directory, and no such location for a tenant without a published name. The
key is public by construction: the name resolves to 127.0.0.1.

Checking the render under the nginx it targets found a pre-existing defect:
`http2 on;` is a 1.25.1+ directive, and avarok2 runs Ubuntu's 1.18, where it
is unknown and the vhost cannot be enabled at all. `listen 443 ssl http2;`
works on both. scripts/test-nginx-vhost.sh renders with and without the
loopback directory and runs nginx -t inside nginx:1.18-alpine and 1.30-alpine
with a throwaway certificate mounted; it runs in CI. Control: the directive
put back is rejected by 1.18.

Stacked on #91.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -220,6 +220,11 @@ jobs:
 
       - name: Tenant provisioning refuses what it must
         run: bash scripts/test-provision-tenant.sh
+      - name: The tenant vhost is accepted by nginx 1.18 and 1.30
+        # avarok2 runs Ubuntu's 1.18; the UI image is on 1.30. `http2 on;` was rejected by the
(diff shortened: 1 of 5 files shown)


tbraun96:

PR #95, opened 2026-09-05
test: run the hosted-UI → loopback-agent path end to end, locally

tbraun96:
Starts the built UI image with a published loopback origin, the agent binary
with a throwaway certificate for local.test, and the Playwright spec that
drives one Chromium at both -- work.test and local.test resolved to
127.0.0.1 by the browser. Controls by environment: an agent that refuses the
page's origin, and an image that publishes no origin, must each fail the
hosted test; both do, and the messages name what the browser saw.

Stacked on #94. The spec lives in citadel-workspace-ui (stacked on #23). CI wiring (build the agent binary in the job that builds the UI image, then run this) is the next step once the queue drains.


tbraun96:

PR #105, opened 2026-09-05
chore: point at the UI that has first-run onboarding

tbraun96:
The parent pinned `citadel-workspaces` at `53defed4`, two commits behind the
UI's master. Those two commits are the whole of production onboarding (UI #19)
and the Lighthouse gate fix (UI #20). Onboarding was written, reviewed and
merged, and then reached nothing: the stack this repo builds, deploys and tests
did not contain it.

```
$ git log --oneline 53defed4..origin/master
8e68c2db fix: the Lighthouse gate measured the runner, not the app (#20)
2a882f69 feat: production-only first-run onboarding, with specs for both paths (#19)
```

### Verified against the running stack, both directions

`http://127.0.0.1:5291/?onboarding=1` → clicking **Create Account** opens
"Before you create an account", naming the two paths and what each needs:

- *Setting up a new workspace* — "You will need the workspace master password
  from the server configuration to become the administrator."
- *Joining a workspace someone else set up* — "You will need its address. You do
  not need the master password, and should not be asked for it."

`http://127.0.0.1:5291/` (dev default) → clicking **Create Account** goes
straight to the wizard, "Connect to a server", Step 1 of 3. No dialog, no extra
interactions.

That second one is the control that matters. `isOnboardingEnabled()` returns
`!isDev` unless `?onboarding=1|0` or a localStorage opt-in overrides it, so the
integration suite's ~90 account creations never see the dialog, while the two
production specs opt in explicitly — one for a new administrator, one for a new
member.

No console errors in either run.


tbraun96:
Superseded. This moved the UI pointer to 48815e3 (first-run onboarding); master's pointer is now 176ee3f, of which 48815e3 is an ancestor, so everything this would bring in is already on master.

https://claude.ai/code/session_01HgeNHXuV6rPNjMXZR2DXsk

PR #112, opened 2026-09-05
feat(agent): TLS on loopback for a hosted UI — name, certificate source, and the wiring

tbraun96:
Reopens the work from #93, rebased onto current master. That PR was closed
while its base moved; the branch survived and this is it, replayed cleanly (one
commit — the release fix it also carried is already on master).

## Why this is the last code gap for the hosted UI

`work.avarok.net` now serves the Citadel UI, and the page dials
`wss://local.avarok.net:12345`. It is served over HTTPS, so the browser refuses
a plain `ws://` upstream — mixed content. The agent must terminate TLS on
loopback.

citadel-agent#60 provides the primitives in the connector (`Transport`,
`acceptor_from_pem`, `HostPolicy`). **The parent binary does not wire them.**
Measured: `citadel-workspace-internal-service --help` has no `--tls-cert`,
`--tls-key` or `--loopback-host`, and its only TLS mention is the `--dangerous`
skip-verify for the *outbound* protocol connection. So there is currently no
way to run an agent the hosted page can reach.

## What it adds

`--loopback-host NAME` is the name the operator points at 127.0.0.1 and holds a
certificate for; the hosted page dials `wss://NAME:PORT`.

The certificate comes from `--tls-cert`/`--tls-key`, or is fetched from
`--loopback-cert-url` (`URL/loopback.pem` and `URL/loopback.key`, https only) at
start and cached `0600` under the data dir, so later starts work offline and a
renewed certificate replaces the cached one. Env equivalents take precedence, as
for every other flag here.

A name without a certificate source is **refused** rather than run plain.
"I set `--loopback-host` and it silently did nothing" is the failure this
avoids, and it is the same shape as the release that shipped a binary which
listened but could not speak WebSocket.

## Dependency

Pins the connector at citadel-agent#60's branch head. That must merge first,
and this pointer bumped to whatever it merges as.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -759,6 +759,7 @@ dependencies = [
  "parking_lot",
  "serde",
  "serde_json",
+ "tokio-rustls",
  "tokio-tungstenite 0.21.0",
  "tungstenite 0.21.0",
  "uuid",
@@ -3926,6 +3927,7 @@ version = "0.23.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c665f33d38cea657d9614f766881e4d510e0eda4239891eea56b4cadcf01801b"
 dependencies = [
+ "log",
  "once_cell",
  "ring",
  "rustls-pki-types",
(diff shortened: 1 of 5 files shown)


tbraun96:

